PR #756, opened 2024-10-30
[DS-442] Adding suppression

racanosmanvs:


src/main/resources/suppressed-cves.xml
```diff
@@ -46,4 +46,17 @@
         <cpe>cpe:/a:apache:jackrabbit</cpe>
         <vulnerabilityName>jackrabbit-spi-2.7.4.jar</vulnerabilityName>
     </suppress>
+
```
jlrhurta, line 49: @jcalcines should we have a look at this code? Also, update the suppress date ?

PR #768, opened 2024-11-08
VS-355 - Move Product Search Widget to bottom of all .com pages

martin-taylor-visitscotland:


martin-taylor-visitscotland:
Had a talk with @jlrhurta - there may be more discussions needed around this ticket.

jlrhurta (approved):

PR #780, opened 2024-11-22
DS-697 Change format of read time and send to front end

jcalcines:
- Created a system for global labels to be delivered independently with fallback to "default" (dot-com) Global label
- Add the read time and publish date to the Delivery API


jlrhurta (approved):

site/components/src/main/java/com/visitscotland/brxm/components/content/PageContentComponent.java
```diff
@@ -188,7 +188,7 @@ private Map<String, String> getGlobalLabels(Locale locale) {
      * @param locale: Locale of the request
      */
     private void addGlobalLabel(Map<String, String> map, String key, Locale locale) {
```
martin-taylor-visitscotland, line 190: Definitely out-of-scope for this PR but just wanted to flag it so that I don't forget it 😄 - I think the parameter names can be clearer, for example `Map<String, String> map`.

site/components/src/main/java/com/visitscotland/brxm/services/ResourceBundleService.java
```diff
@@ -55,12 +55,14 @@ public String getResourceBundle(String bundleName, String key, Locale locale){
...
 
-        if (properties.getSiteId().isEmpty()){
-            return getResourceBundle(bundleName, key, locale, false);
-        } else {
-            return getResourceBundle(properties.getSiteId() + "." + bundleName, key, locale, false);
+        if (!properties.getSiteId().isEmpty()){
+            String value = getResourceBundle(properties.getSiteId() + "." + bundleName, key, locale, true);
+            if (value != null) {
+                return value;
+            }
```
martin-taylor-visitscotland, line 62: I think this method can be refactored using the `Optional<T>` framework, perhaps something like:

```java
public String getSiteResourceBundle(String bundleName, String key, Locale locale){
    final Optional<String> resourceValue = Optional.ofNullable(
        getResourceBundle(properties.getSiteId() + "." + bundleName, key, locale, true)
    );

    return resourceValue.orElse(getResourceBundle(bundleName, key, locale, false));
}
```

jcalcines: Hey, I am not too keen on the Optional framework as I think it makes the code less readable.

But I am happy to have a conversation about it, to see if we are missing something :)

You also forgot to check `properties.getSiteId().isEmpty()` which might make the code longer than the original one ;)

martin-taylor-visitscotland: Ello! Sure thing, thanks for the response 😄 - happy to have a discussion around this at some point, as for now - no problem!

martin-taylor-visitscotland (approved):
Great work! 🥳

site/components/src/main/java/com/visitscotland/brxm/components/content/PageContentComponent.java
jcalcines: A developer choosing a bad variable name? 😁 Yeah, something to review :)

jlrhurta (approved):

PR #781, opened 2024-11-22
DS-719 SE - BSH pages adding an extra space to the content type 

jcalcines:
I have converted the theme field to a dynamic DropDown

If we merge this we need update all pagesdocument  in production including the error pages


martin-taylor-visitscotland (approved):
Looks good to me!

jlrhurta (approved):

PR #785, opened 2024-11-25
DS-440-CTA-BANNER - Fix when `Anchor` value is not present

martin-taylor-visitscotland:
As highlighted [in the ticket](https://visitscotland.atlassian.net/browse/DS-440) by Corina, it is expected that when the `anchor` is not present - it should fallback to the module title. This PR introduces these changes.

* Also fixes trailing hyphens, e.g. `Hello world?` becomes `hello-world` and not `hello-world-`.
* Added 100% unit test coverage for `AnchorFormatter.java`.


martin-taylor-visitscotland:
Now that I think about it @jcalcines @jlrhurta - example Corina provided [in the ticket](https://visitscotland.atlassian.net/browse/DS-440) shows the anchor being rewritten, for example:

* `My Module Title` -> becomes -> `my-module-title`

Is this expected behavior?

jlrhurta (approved):

site/components/src/main/java/com/visitscotland/brxm/utils/AnchorFormatter.java
```diff
@@ -0,0 +1,28 @@
...
+            return formatString(fallback.get());
+        }
+
+        return anchor;
+    }
+
+    private String formatString(String string) {
+        return string
+            .replaceAll("[^a-zA-Z0-9]+", "-")
+            .replaceAll("-+$", "")
```
jcalcines, line 25: Just double-checking. This is removing dashes at the end, right?

Also if you lowercase at the beginning you can reduce the complexity of the regex to [^a-z0-9]+

jcalcines (approved):

jcalcines (approved):

jlrhurta (approved):

PR #791, opened 2024-12-03
DS-695 Create External Link Drop-Down Options File

jcalcines:


martin-taylor-visitscotland (approved):

repository-data/application/src/main/resources/hcm-config/configuration/translations/cms.yaml
```diff
@@ -75,9 +75,11 @@ definitions:
         and featured items can not be mixed - pick 2 featured items or 1 featured
         event.
       non-empty-spaces: Only characters and number are allowed, no spaces.
-      visitscotland:article-heading-validator: Each paragraph should start with an Heading
-        for the current layout
+      visitscotland:article-heading-validator: Each paragraph should start with an
+        Heading for the current layout
```
martin-taylor-visitscotland, line 79: Very small point (grammatical) - I believe this should be:

```
Each paragraph should start with a heading for the current layout.
```

site/components/src/main/java/com/visitscotland/brxm/hippobeans/SharedLinkBSH.java
```diff
@@ -6,6 +6,9 @@
 @HippoEssentialsGenerated(internalName = "visitscotland:SharedLinkBSH")
 @Node(jcrType = "visitscotland:SharedLinkBSH")
 public class SharedLinkBSH extends SharedLink {
+
+    public static final String WEBSITE_TYPE = "visitscotland:websiteType";
+
```
martin-taylor-visitscotland, line 11: Just out of curiosity, where will this be used? 😄 

The reason I mention is that I try to typically follow a principle called [YAGNI (You Aren't Gonna Need It)](https://www.techtarget.com/whatis/definition/You-arent-gonna-need-it) that advises against implementing features until they are actually required, promoting simplicity, reducing development time, and minimising maintenance effort by avoiding unnecessary code.

Just thought I'd highlight, definitely not precious about this though - especially if you can foresee a use-case 😄

doc/how-to/commit-yaml-files.md
```diff
@@ -0,0 +1,16 @@
+# Working with yaml files in BloomReach
+
+# Table of Content
+1.
```
martin-taylor-visitscotland, line 4: A few points to raise around the Table of Content 😄:

* It may be missing some content (only mentions `1.`)
* I believe this should be plural - i.e. `Table of Contents`

cms/src/main/java/com/visitscotland/brxm/validator/SharedLinkExternalValidator.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.Set;
+
+/**
+ * jcr:Name = visitscotland:shared-link-bsh-external
+ *
+ * @author jose.calcines
+ */
+public class SharedLinkExternalValidator implements Validator<Node> {
+
+    private static final Logger logger = LoggerFactory.getLogger(LinkValidator.class);
```
martin-taylor-visitscotland, line 28: Small point 😄 typically the convention in Java is that static constants should be capitalised!

```suggestion
    private static final Logger LOGGER = LoggerFactory.getLogger(LinkValidator.class);
```

What do you think? 😄

repository-data/application/src/main/resources/hcm-config/configuration/translations/cms.yaml
jcalcines: Hehe, Copy/Paste method went wrong, thank you for spotting that!

doc/how-to/commit-yaml-files.md
jcalcines: I guess it can be singular if it only has one item 😄 . Just kidding! I shouldn't have added that Table of Contents section just yet. I will remove it for the time being.

cms/src/main/java/com/visitscotland/brxm/validator/SharedLinkExternalValidator.java
jcalcines: The eternal question about Logger being a constant, right 😄?

I would lean toward consistency here, so given that the rest of the codebase is using logger as lower case I would stick with that.

About the observation itself. I have been in both sides on the fence and technically the logger is not inmutable therefore it cannot be classified as pure constant that's why it is set in lowercase (If you look at the [documentation](https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names) you can see that they also classify it as not constant.

However, I see your point, I personally agree with it. I preffer to use upper case so I can denote the intention of inmutability.

I hope this made sense

martin-taylor-visitscotland: This was very valuable, thank you! 😁

site/components/src/main/java/com/visitscotland/brxm/hippobeans/SharedLinkBSH.java
jcalcines: Note: I created the constant but I did not use it. It was meant to be used in the SharedLinkValidator class. That's now fixed

cms/src/main/java/com/visitscotland/brxm/validator/SharedLinkExternalValidator.java
```diff
@@ -0,0 +1,47 @@
...
+        try {
+                for (NodeIterator it = node.getNodes(SharedLink.LINK_TYPES); it.hasNext(); ) {
+                    Node link = it.nextNode();
+                    if (link.isNodeType("visitscotland:ExternalLink")) {
+                        if (node.hasProperty("visitscotland:websiteType") && node.getProperty("visitscotland:websiteType").getString().isEmpty()){
+                            return Optional.of(context.createViolation());
+                        }
+                    }
+                }
+            return Optional.empty();
```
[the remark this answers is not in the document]
jcalcines: This should be resolved now

cms/src/main/java/com/visitscotland/brxm/validator/SharedLinkExternalValidator.java
```diff
@@ -0,0 +1,47 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(LinkValidator.class);
+
+
+
+    public Optional<Violation> validate(final ValidationContext context, final Node node) {
+        try {
+                for (NodeIterator it = node.getNodes(SharedLink.LINK_TYPES); it.hasNext(); ) {
+                    Node link = it.nextNode();
+                    if (link.isNodeType("visitscotland:ExternalLink")) {
+                        if (node.hasProperty("visitscotland:websiteType") && node.getProperty("visitscotland:websiteType").getString().isEmpty()){
```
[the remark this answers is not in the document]
jcalcines: This is fixed now :)

martin-taylor-visitscotland (approved):

PR #793, opened 2024-12-05
BRXM Personalisation PoC - Version 1

martin-taylor-visitscotland:


site/components/src/main/java/com/visitscotland/brxm/personalisation/PersonalisationHandler.java
martin-taylor-visitscotland, line 1: Note to self, re-think this and migrate away from using `static`. This should be thread-safe, there is no shared mutable state. Tomcat typically operates in a thread-per-request model, so each request should be executed on a separate thread.

Though typically I'd like to avoid the use of static where necessary, possibly a candidate for a bean?

site/components/src/main/java/com/visitscotland/brxm/personalisation/PersonalisationService.java
```diff
@@ -0,0 +1,39 @@
...
+
+import org.springframework.stereotype.Component;
+
+import java.util.stream.Collectors;
+import java.util.Optional;
+import java.util.List;
+
+@Component
+public class PersonalisationService {
+    private static final String PERSONALIZATION_JCR_TYPE = "visitscotland:personalization";
```
martin-taylor-visitscotland, line 17:
```suggestion
    private static final String PERSONALIZATION_JCR_TYPE = "visitscotland:Personalization";
```

site/components/src/main/java/com/visitscotland/brxm/personalisation/PersonalisationService.java
```diff
@@ -0,0 +1,44 @@
...
+
+    private boolean matchesVisitorContextCountry(Personalization personalization) {
+        return getVisitorContextFromRequest()
+            .filter(visitorContext -> personalization.getCountry().equals(visitorContext.getCountry()))
+            .isPresent();
+    }
+
+    private Optional<VisitorContext> getVisitorContextFromRequest() {
+        final HstRequestContext requestContext = RequestContextProvider.get();
+        final VisitorContext visitorContext = (VisitorContext) requestContext.getAttribute("visitorContext");
```
jcalcines, line 40: This is a preference comment, so you can freely ignore it.

I know that the way you have set that is kind of standard but I wonder if you need to define both variable. I would have done this instead: 
`        final VisitorContext visitorContext = (VisitorContext) RequestContextProvider.get().getAttribute("visitorContext");`
        
In my head the advantages are:
- You need one less import as you really don't need anything directly from the HstRequestContext object.
- It is one less line of code you need to understand. A glance to the code says that only the visitorContext is useful for your piece of code.
- You don't need to allocate memory for the pointer to requestContext which will be disposed after the method which makes it technically a tiny bit more resource friendly 

Again, this is just a matter of preference and your approach is more standard than mine. It is just a practice I usually do and I wanted to see your thoughts about it

site/components/src/test/java/com/visitscotland/brxm/personalisation/PersonalisationServiceTest.java
```diff
@@ -0,0 +1,145 @@
...
+    private static final String UNITED_KINGDOM = "UK";
+
+    private static final VisitorContext VISITOR_CONTEXT_UK = new VisitorContext(UNITED_KINGDOM);
+
+    public PersonalisationServiceTest() {
+        this.personalisationService = new PersonalisationService();
+    }
+
+    @Test
+    void When_GetPersonalisedVariants_With_NoPersonalizationCompounds_Expect_EmptyList() {
```
jcalcines, line 46: Can you include the annotation @DisplayName explaining the purpose of the test, instead of using a long name for the method?

Also, I think it is worth for us to discuss name convention for the methods 😄

site/components/src/test/java/com/visitscotland/brxm/personalisation/PersonalisationServiceTest.java
```diff
@@ -0,0 +1,145 @@
...
+
+        Assertions.assertTrue(result.isEmpty());
+    }
+
+    @Test
+    void When_GetPersonalisedVariants_With_PersonalizationCompoundsNotMatchingVisitorContext_Expect_EmptyList() {
+        try (MockedStatic<RequestContextProvider> requestContextProviderMock = mockStatic(RequestContextProvider.class)) {
+            final Personalization personalisedVariant = mock(Personalization.class);
+
+            requestContextProviderMock
```
jcalcines, line 64: Neat code indentation. I will borrow it

site/components/src/test/java/com/visitscotland/brxm/personalisation/PersonalisationServiceTest.java
```diff
@@ -0,0 +1,145 @@
...
+import static org.mockito.Mockito.eq;
+
+@ExtendWith(MockitoExtension.class)
+class PersonalisationServiceTest {
+    @Mock private MockHippoBean hippoBean;
+    @Mock private MockHstRequestContext requestContext;
+
+    private final PersonalisationService personalisationService;
+
+    private static final String PERSONALIZATION_JCR_TYPE = "visitscotland:Personalization";
```
jcalcines, line 34: Maybe you can use the following?
private static final String PERSONALIZATION_JCR_TYPE = Personalization.PRIMARY_TYPE

site/components/src/test/java/com/visitscotland/brxm/personalisation/PersonalisationServiceTest.java
```diff
@@ -0,0 +1,145 @@
...
+    private static final String VISITOR_CONTEXT_ATTRIBUTE = "visitorContext";
```
jcalcines, line 35: I have seen that text hard-coded somewhere in the main classes. Is it ok to move the constant there and reuse it from the class?

jcalcines (approved):

site/components/src/test/java/com/visitscotland/brxm/personalisation/PersonalisationServiceTest.java
martin-taylor-visitscotland: Hey Jose,

The reason I don't like to mix production code with tests:

* Test Independence: Tests should be independent of the production code to ensure they are testing the behavior and not the implementation. Using the same constants can lead to false positives, where tests pass because they are using the same values as the production code, rather than verifying the actual behavior.
* Clarity: Keeping test data separate from production code can make your tests clearer and easier to understand. It helps in identifying what the test is verifying without having to cross-reference the production code.

What do you think?

site/components/src/main/java/com/visitscotland/brxm/personalisation/CountryValidator.java
```diff
@@ -0,0 +1,8 @@
+package com.visitscotland.brxm.personalisation;
+
+class CountryValidator implements Validator<String> {
+    @Override
+    public boolean isValid(String country) {
```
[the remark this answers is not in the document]
martin-taylor-visitscotland: Absolutely, great spot! :)

site/components/src/main/java/com/visitscotland/brxm/hippobeans/Personalization.java
```diff
@@ -8,13 +8,15 @@
 @HippoEssentialsGenerated(internalName = "visitscotland:Personalization")
 @Node(jcrType = "visitscotland:Personalization")
 public class Personalization extends HippoCompound {
-    @HippoEssentialsGenerated(internalName = "visitscotland:id")
-    public String getId() {
-        return getSingleProperty("visitscotland:id");
-    }
+    public static final String PERSONALIZATION_JCR_TYPE = "visitscotland:Personalization";
```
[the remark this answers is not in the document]
martin-taylor-visitscotland: This is an interesting one, happy to do this in terms of consistency for the meantime but personally I think 'PRIMARY_TYPE' can be a tad confusing - depending on the context you read it.

Maybe we can have a quick discussion in the technical catch up, but I personally think this constant can be renamed a bit better for the purpose of clarity!

I often ask myself when I read those constants, "what type?".

Even 'PRIMARY_JCR_TYPE' could add a bit more value in my eyes but keeping it explicit ensures it will always make sense, statically imported or not.

What do you think?

site/components/src/main/java/com/visitscotland/brxm/personalisation/PersonalisationService.java
```diff
@@ -0,0 +1,44 @@
...
+import org.hippoecm.hst.core.request.HstRequestContext;
+
+import org.springframework.stereotype.Component;
+
+import java.util.stream.Collectors;
+import java.util.Optional;
+import java.util.Objects;
+import java.util.List;
+
+import static com.visitscotland.brxm.hippobeans.Personalization.PERSONALIZATION_JCR_TYPE;
```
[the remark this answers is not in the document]
martin-taylor-visitscotland: Happy to revert in terms of organisational preference, though I do think it could be a good discussion point!

I've nothing against static imports, if used properly I think they can improve code readability - though just like anything if used improperly it can negatively impact it :D

site/components/src/main/java/com/visitscotland/brxm/hippobeans/Personalization.java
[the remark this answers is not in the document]
jcalcines: Yeah, This comment was on the back of my assumption about not using static imports, so the code would read like `Personalization.PRIMARY_TYPE`. Otherwise, my comment here makes no sense and your name is more descriptive.

Maybe we can touch this point when we talk about static imports

site/components/src/test/java/com/visitscotland/brxm/personalisation/PersonalisationServiceTest.java
martin-taylor-visitscotland: It's open source 😉 haha!

site/components/src/test/java/com/visitscotland/brxm/personalisation/PersonalisationServiceTest.java
jcalcines: Great answer! I understand your point of view and now I have mixed thoughts about this.

However, My main point about this is that the definition of the constant comes from somewhere else and it is not something you made up for this test or is coincidentally matching some value. Essentially, in my eyes, I see this constants like something fixed and part of the original requirement.

Those constants won't change, but in case they do, you might want your unit test to reflect that change automatically.

Does this makes sense?

site/components/src/test/java/com/visitscotland/brxm/personalisation/PersonalisationServiceTest.java
martin-taylor-visitscotland: This is yet another good discussion point, absolutely happy to do this as it's the way of working that's been established at VS I've noticed! There are a few points I'd like to raise around this:

* Readability and Maintainability: Clean and descriptive test method names make it easier to understand what the test is doing at a glance. This is especially useful when reading the code in an IDE or a code review
* Consistency: Using a consistent naming convention for test methods helps maintain uniformity across the test suite. This can make it easier for new developers to understand the tests and for existing developers to maintain them1
* IDE and Tooling Support: Many IDEs and build tools display test method names by default. Having clear and descriptive method names ensures that the test reports and logs are immediately understandable without needing to look at the @DisplayName annotation.
* Avoiding Redundancy: If the test method names are already clear and descriptive, using @DisplayName can be redundant. It’s better to invest time in writing good method names rather than relying on annotations to explain what the test does.

martin-taylor-visitscotland: Happy to do this :)

site/components/src/main/java/com/visitscotland/brxm/personalisation/PersonalisationService.java
martin-taylor-visitscotland: This is an interesting point :)

Given that the performance impact will be microscopic and the GC will kick in after the scope of the method is executed, I think, personally, I favor the more explicit declarations over one-liners! But again, in context a one-liner may be perfectly readable and OK to do!

What do you think?

jcalcines: That's fair, this is just a matter of preference. I'm sharing my preferred way of working.

To be honest there is not that much of a difference :)

site/components/src/main/java/com/visitscotland/brxm/hippobeans/Personalization.java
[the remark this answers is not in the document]
martin-taylor-visitscotland: This makes sense, though `Personalization.PRIMARY_TYPE` still in my opinion may seem quite confusing (depending on context) - what about something like `Personalization.JCR_TYPE`?

I think this way the intention and purpose of the constant is clear!

Again, happy to wait and have a winder discussion. Also happy for me to also rename to `Personalization.PRIMARY_TYPE` for now if this is what you'd like me to do :D

site/components/src/test/java/com/visitscotland/brxm/personalisation/PersonalisationServiceTest.java
martin-taylor-visitscotland: I see your point and totally makes sense :)

I still have a counter point to this though, while constants may come from the original requirements, tests should still be independent to ensure they are verifying the behavior and not just the implementation. If the constant changes, it should be a deliberate decision to update the test data to reflect the new behavior, rather than having tests pass automatically without scrutiny.

What do you think? We can always have this as another point to discuss on the technical discussion! 😄

repository-data/application/src/main/resources/hcm-config/namespaces/visitscotland/Personalization.yaml
```diff
@@ -33,6 +33,30 @@ definitions:
             hipposysedit:primary: false
             hipposysedit:type: hippo:mirror
             hipposysedit:validators: [required]
+        /hipposysedit:nodetype[2]:
```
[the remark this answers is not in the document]
martin-taylor-visitscotland: This is an interesting point, let me check and get back to you! Amazing spot!

site/components/src/main/java/com/visitscotland/brxm/hippobeans/Personalization.java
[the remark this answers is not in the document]
jcalcines: Happy to discuss, but also `Personalization.JCR_TYPE` sounds great. I took the term Primary Type from BloomReach name convention, but I JCR_TYPE feels more accurate.

site/components/src/main/java/com/visitscotland/brxm/utils/PageTemplateBuilder.java
```diff
@@ -124,7 +129,14 @@ public void addModules(HstRequest request, String location) {
 
     private void addModule(HstRequest request, PageConfiguration page, BaseDocument item, String location){
         if (item instanceof Megalinks) {
-            processMegalinks(request, page, (Megalinks) item);
+            // TODO FOR DEMO ONLY, LOGIC WILL BE POLISHED
```
jcalcines, line 132: Remember to Tidy up this section :)

jcalcines (approved):
Hey Martin,

I think you can progress with the comments now we have validated that version 1 is what we want to move forward.

One thing I would like you to do as part of this PR is to move the personalization package to com.visitscotland.brxm.services.personalization. I think it is ok to keep all classes in the same package as they are all very specific to personalization

PR #797, opened 2024-12-09
FEAT DS-744 (Cludo search) Adding cludo search page results

jlrhurta:


martin-taylor-visitscotland (approved):

jcalcines (approved):
@jlrhurta I am missing the part when you tell the front-end that the page is a Cludo Search. Is that currently comming from GeneralContentComponent?

PR #806, opened 2024-12-20
DS-835 - Allow SVG Images in the CMS

jcalcines:


martin-taylor-visitscotland:
Hey @jcalcines - where did the requirement come from to allow `SVG` images to be uploaded from CMS users? Always a little hesitant with allowing `SVG` due to the security risks around these file-types (a good article https://www.fortinet.com/blog/threat-research/scalable-vector-graphics-attack-surface-anatomy).

When a user uploads an SVG graphic - is there any sanitation done?

What do you think? 😄

jcalcines:
@martin-taylor-visitscotland 

If I understand the issue correctly, If we were to host the SVGs, the problem is that we might end up serving malicious images, so If the system were used only to upload internally generated SVGs we should be all OK, right?
 
I believe that the policy on VisitScotland is to only use internal images, so I believe we should be safe, and the likelihood is that we don't have space to change this requirement right now. However, I will reach Nicola and Gavin to make them aware of this issue and to design a potential solution.

jcalcines:
@martin-taylor-visitscotland , @jlhurta & @gavin-vs, we have agreed to allow SVG for the time being on the CMS. 

The risk is for the content team to upload untrusted SVG into the CMS. So the solution is 
1. To make aware to the content team about this issue
2. To develop (later on) a sanitizing service for SVGs, so whenever we need an SVG from the CMS, we use sanitized versions of them.

cc @jameshaddrillvs

PR #807, opened 2024-12-24
Create Unit Test for CMS Validator

jcalcines:
Hello Guys,

Happy new year!

I have completed the unit test with decent coverage on the code (82%).

Ihave also noticed that some constants were note created of missplaced so I had to alter classes other than the validator

I have also flagged some issues with TODO with questions as it seems like some validations around maps, might not be in use.


site/components/src/main/java/com/visitscotland/brxm/hippobeans/ArticleStyledSection.java
```diff
@@ -4,12 +4,14 @@
...
  * TODO: Beanwriter: Failed to create getter for node type: hippo:compound
  */
 @HippoEssentialsGenerated(internalName = "visitscotland:ArticleStyledSection")
 @Node(jcrType = "visitscotland:ArticleStyledSection")
 public class ArticleStyledSection extends ArticleSection {
+    public static final String HEADING = "visitscotland:heading";
+
     @HippoEssentialsGenerated(internalName = "visitscotland:heading")
     public String getHeading() {
         return getSingleProperty("visitscotland:heading");
```
jlrhurta, line 17: Should not you use the constant HEADING in line 17?

jcalcines: We could use it in line 17 but not in line 15, that's why we don't normally use them from the hippoBean.

That constant is intended for use in other classes so you can denote a dependency between this model an where it is use on (In this case the validator)

site/components/src/main/java/com/visitscotland/brxm/hippobeans/MapModule.java
```diff
@@ -7,12 +7,16 @@
...
-/** 
+/**
  * TODO: Beanwriter: Failed to create getter for node type: hippo:compound
  */
 @HippoEssentialsGenerated(internalName = "visitscotland:MapModule")
 @Node(jcrType = "visitscotland:MapModule")
 public class MapModule extends BaseDocument {
+
+    public static final String MAP_KEYS = "hippotaxonomy:keys";
+    public static final String MAP_TYPE = "visitscotland:mapType";
```
martin-taylor-visitscotland, line 18: I think these variable names can be a tad clearer, what do you think? 😄 

Just highlighting that it's some JCR related property could help with readability.

site/components/src/main/java/com/visitscotland/brxm/hippobeans/MapModule.java
```diff
@@ -7,12 +7,16 @@
 import java.util.Calendar;
 import java.util.List;
 
-/** 
+/**
  * TODO: Beanwriter: Failed to create getter for node type: hippo:compound
  */
```
martin-taylor-visitscotland, line 12: What does this mean? Is this something that needs to be addressed?

cms/src/test/java/com/visitscotland/brxm/validator/SharedLinkExternalValidatorTest.java
martin-taylor-visitscotland, line 1: This implementation looks good! I just wanted to point out that there are no [verifications](https://site.mockito.org/javadoc/current/org/mockito/Mockito.html#verification) in the test. Adding verifications could help ensure that expected interactions with dependencies actually occur, rather than just relying on state assertions.

Would it be possible to introduce some verify() calls? This would help confirm that:
* Dependencies are invoked the expected number of times (or not invoked at all).
* The correct arguments are passed to collaborators.
* The overall behavior aligns with expectations, preventing silent regressions.

This would make the test more robust by validating both state and behavior. Let me know what you think!

cms/src/test/java/com/visitscotland/brxm/validator/NumberValidatorTest.java
```diff
@@ -0,0 +1,60 @@
...
+        // Execute
+        Optional<Violation> result = validator.validate(context, value);
+
+        // Verify
+        assertEquals(isValid, result.isEmpty());
+    }
+
+    @Test
+    @DisplayName("Should return a violation when the value is empty")
+    void testValidate_EmptyValue() {
```
martin-taylor-visitscotland, line 47: I think the test names could be a bit clearer. Right now, they follow the `methodName_StateUnderTest` pattern, but I’d recommend using` methodName_StateUnderTest_ExpectedResult` instead.

Even though there's a `@DisplayName`, it's generally best practice to make test method names self-explanatory. A well-structured name improves readability, making it immediately clear what’s being tested and what the expected outcome is. This helps when quickly scanning tests or debugging failures.

Would do you think 😄?

cms/src/test/java/com/visitscotland/brxm/validator/NumberValidatorTest.java
martin-taylor-visitscotland, line 1: Just want to highlight [this comment](https://github.com/visitscotland/dot-com-brxm/pull/807/files#r1936973486) again! 😄

cms/src/test/java/com/visitscotland/brxm/validator/MapPageValidatorTest.java
martin-taylor-visitscotland, line 1: Just want to highlight [this comment](https://github.com/visitscotland/dot-com-brxm/pull/807/files#r1936973486) again! 😄

cms/src/main/java/com/visitscotland/brxm/validator/ContentListHeadingValidator.java
```diff
@@ -58,7 +59,7 @@ private boolean hasValidHeading(final Node node) throws RepositoryException {
         if (themes.contains(node.getProperty(Article.THEME).getValue().getString())){
             for (NodeIterator it = node.getNodes(Article.PARAGRAPH); it.hasNext(); ) {
                 Node section = it.nextNode();
-                if (Contract.isEmpty(section.getProperty("visitscotland:heading").getString())){
+                if (Contract.isEmpty(section.getProperty(ArticleStyledSection.HEADING).getString())){
```
jlrhurta, line 62: Is this only used for ArticleStyled? dont we use it for articles for VS or BSH?

martin-taylor-visitscotland (approved):
Small comments again that can be discussed, the only comment I really think is worth highlighting at the moment is the one around verifications!

cms/src/main/java/com/visitscotland/brxm/validator/ContentListHeadingValidator.java
jcalcines: No, at the moment this heading is only available for the Article Styled document type

cms/src/test/java/com/visitscotland/brxm/validator/NumberValidatorTest.java
jcalcines: That's a good suggestion but we probably want to define the test process independently to this PR. At the moment I think it is better If we stick with consistency

I don't mind making methods names a bit more descriptive, but I feel that in term of readability, the description under DisplayName is more readable and it states clear the high level purpose of the unit and it can also reference to the original ticket

cms/src/test/java/com/visitscotland/brxm/validator/SharedLinkExternalValidatorTest.java
jcalcines: This is a good point!  We might need to review the way we design the unit tests.

Traditionally, using 'Spy' to verify outputs, used to make the unit test more flaky specially when the requirements change as the class you are testing is no longer itself but a proxy and the debugger cannot match the code in runtime.

However, things might have improved since we got in those situations. I am happy to discuss this change of approach

site/components/src/main/java/com/visitscotland/brxm/hippobeans/MapModule.java
jcalcines: Hello, I was just moving the constants to their origin for clarity. I'm not sure what names we could use instead

site/components/src/main/java/com/visitscotland/brxm/hippobeans/MapModule.java
jcalcines: This happens when BloomReach is not able to generate getter or setter for some of the properties, the todo is for developers to add the field manually.

The action for this one might be just to remove the comment. I will check this as part of https://visitscotland.atlassian.net/browse/BSHUB-614

jcalcines:
> Small comments again that can be discussed, the only comment (from me) I really think is worth highlighting at the moment is the one around verifications!

Hi, @DukesColleague, I have answered your comments. We essentially have bad experiences with `Spy` classes. This could be something we might need to review in the technical forums

DukesColleague (approved):
Lovely to see some coverage be added 😄

PR #816, opened 2025-01-13
DS-806-BACK-END-HELPERS - Migrate FTL helper methods

DukesColleague:
This should now be complete as labels are retrievable within the `ResourceBundleService.java` Spring bean, a refactoring pull-request should follow at some point as I have identified a few improvements around this.

For now, this PR migrates the following into static utility methods:
* `addParameter` has been relocated to `utils` / `UrlParameterAppender.java`
* `escapeJSON` has been relocated to `utils` / `JsonEscaper.java`
* `getUrl` has been relocated to `utils` / `UrlResolver.java`

These other helper methods can be ignored:
* `label`
* `optionalLabel`
* `property`
* `labelFallback`
* `productSearch`

### DIFF
--- a/site/components/src/main/java/com/visitscotland/brxm/utils/JsonEscaper.java
+++ b/site/components/src/main/java/com/visitscotland/brxm/utils/JsonEscaper.java
@@ -1,0 +1,20 @@
+package com.visitscotland.brxm.utils;
+
+public final class JsonEscaper {
+    private JsonEscaper() { }
+
+    public static String escape(String original, boolean isJsonObject) {
+        String escaped = original.replace("'", "\\'");
+        escaped = isJsonObject ? escapeJsonObject(escaped) : escapeQuotes(escaped);
+
+        return escaped;
+    }
+
+    private static String escapeJsonObject(final String json) {
+        return json.replace("\"", "'");
+    }
+
+    private static String escapeQuotes(final String json) {
+        return json.replace("\"", "&quot;");
+    }
+}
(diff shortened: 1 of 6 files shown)


jcalcines (commented):
Hey Martin, 

I feel a bit bad about this, I don't think captured the spirit of the ticket properly. Technically speaking the code is ok. I am a bit used to have a method doing the work over a class doing the work but I understand the principle (SRD) behind this so I like the approach.

However, this ticket was mostly about understand if these helpers where needed for the Delivery API implementation. I believe most of this operations will be handled by the front-end side otherwise I would be expecting this PR use include the usage of this classes.

I'm not suggesting that this classes are not needed but flagging that we need a conversation and/or some extra steps before this is ready to be merged.

martin-taylor-visitscotland:
> Hey Martin,
> 
> I feel a bit bad about this, I don't think captured the spirit of the ticket properly. Technically speaking the code is ok. I am a bit used to have a method doing the work over a class doing the work but I understand the principle (SRD) behind this so I like the approach.
> 
> However, this ticket was mostly about understand if these helpers where needed for the Delivery API implementation. I believe most of this operations will be handled by the front-end side otherwise I would be expecting this PR use include the usage of this classes.
> 
> I'm not suggesting that this classes are not needed but flagging that we need a conversation and/or some extra steps before this is ready to be merged.

Hey @jcalcines,

Very valid points! Now that I think about it - I do think these methods would be better suited to the FE!

PR #819, opened 2025-01-15
DS-562 - Add data for Related Links

jcalcines:
This ticket is supposed to fullfil the following acceptance criteria:
1. External links can be included.  They do not include read time.  The content type will always be ‘Partner website’
2. Download links can be included.  They will include the filesize and format

The first point won't be fully completed by the backend but It will depends on the links selected by the content team.

JIRA: https://visitscotland.atlassian.net/browse/DS-562

### DIFF
--- a/site/components/src/main/java/com/visitscotland/brxm/services/LinkService.java
+++ b/site/components/src/main/java/com/visitscotland/brxm/services/LinkService.java
@@ -497,7 +497,7 @@ public class LinkService {
      * @param sharedLink  SharedLink document that contains extra information
      * @param module      Module to feed with any possible issue found while creating the page.
      * @param locale      Language for the label
-     * @param addCategory wether or not the category field is populated.
+     * @param addCategory whether the category field is populated.
      */
     private EnhancedLink enhancedLinkFromSharedLink(SharedLink sharedLink, Module<?> module, Locale locale, boolean addCategory) {
         EnhancedLink link = new EnhancedLink();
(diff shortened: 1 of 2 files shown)


jlrhurta (approved):

site/components/src/main/java/com/visitscotland/brxm/services/LinkService.java
```diff
@@ -513,6 +513,9 @@ private EnhancedLink enhancedLinkFromSharedLink(SharedLink sharedLink, Module<?>
         if (sharedLink instanceof SharedLinkBSH) {
             SharedLinkBSH sharedLinkBSH = (SharedLinkBSH) sharedLink;
             link.setSource(sharedLinkBSH.getSource());
+            if (!Contract.isEmpty(sharedLinkBSH.getWebsiteType())) {
```
martin-taylor-visitscotland, line 516: Just a more broader question, I have been thinking about the `Contract` class - there are a few utilities within that class that Vanilla Java is able to handle. Would it make sense to delegate the responsibility to Vanilla Java so we don't need to worry about the maintenance of these utilities - and keep the more specific use-cases? 😄

site/components/src/main/java/com/visitscotland/brxm/services/LinkService.java
```diff
@@ -513,6 +513,9 @@ private EnhancedLink enhancedLinkFromSharedLink(SharedLink sharedLink, Module<?>
...
+                link.setCategory(utils.getValueMap("bsh-external-links").get(sharedLinkBSH.getWebsiteType()));
```
martin-taylor-visitscotland, line 517: Verrrry small point, out of scope - and totally aside from your PR (sorry) just wanted to note that I think that the readability of these can be improved by renaming `utils` to `hippoUtilities` and `getValueMap` to `getValueListMap()`.

So that it reads:
`hippoUtils.getValueList()` - as `ValueList` is the proper DSL for Hippo!

site/components/src/main/java/com/visitscotland/brxm/services/LinkService.java
jcalcines: Absolutelly, I think it is a fantastic idea, as long as they fulfil the same purpose. It is always better to use Vanilla Java when possible.

site/components/src/main/java/com/visitscotland/brxm/services/LinkService.java
jcalcines: Also a good point! I struggled a bit to find the method because It wasn't explicit about "ValueList". 

My guess is that the original Idea was to convert the list to a map and that's why we chose that name

site/components/src/main/java/com/visitscotland/brxm/services/LinkService.java
jcalcines: @martin-taylor-visitscotland , out of curiosity. Which vanilla function would you use?

PR #827, opened 2025-01-21
DS-562 - Replace keys with Texts for BSH Categories

jcalcines:
JIRA: https://visitscotland.atlassian.net/browse/DS-562

### DIFF
--- a/site/components/src/main/java/com/visitscotland/brxm/services/LinkService.java
+++ b/site/components/src/main/java/com/visitscotland/brxm/services/LinkService.java
@@ -651,26 +651,31 @@ public class LinkService {
      * @param sectors sectors selected
      * @param regions regions selected
      */
-    private void setBSHFields (EnhancedLink link, String contentType, String[] sectors, String skill, String[] topics, String[] regions){
-        link.setContentType(utils.getValueMap("bsh-content-types").get(contentType));
-        link.setSkillLevel(utils.getValueMap("bsh-skill-levels").get(skill));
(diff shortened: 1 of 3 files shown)


site/components/src/main/java/com/visitscotland/brxm/services/LinkService.java
```diff
@@ -653,18 +651,31 @@ private String getYoutubeId(String url) {
...
+                    .map(topic -> utils.getValueMap(TOPICS_VALUE_LIST).get(topic))
+                    .collect(Collectors.toList()));
         }
         if (regions != null) {
             link.setRegion(List.of(regions));
+            link.setRegion(Arrays.stream(regions)
+                    .map(region -> utils.getValueMap(REGIONS_VALUE_LIST).get(region))
+                    .collect(Collectors.toList()));
         }
     }
```
martin-taylor-visitscotland, line 680: Quite a large method I think can be broken down into their own responsibilities - I see potentially a couple other methods such as `setContentType`, `setSkillLevel`, `setSectors`, `setTopics` and `setRegions`.

What do you think? :)

jlrhurta: I agree but also I think this could be added as TODO and refactor inmediately after merging to avoid conflicts with sandbox-development data?

jcalcines: I'm not keen of fractioning this method as the only thing is doing is populating fields from the model. 

However, I understand your point I am going to try something else that might improve readability

jlrhurta (approved):

martin-taylor-visitscotland (approved):

PR #831, opened 2025-01-24
FEAT DS-899 (Events BSH) Events listing content modelling

jlrhurta:


martin-taylor-visitscotland (approved):
As talked about on Teams, I agree that splitting the labels shouldn't block DS-899 😄

jcalcines (approved):

PR #837, opened 2025-01-27
DS-964 - Allow links to Internal assets

jcalcines:
Hello Guys,

This is what I have build for allowing links to CMS assets

image

I don't entirely agree with this approach to allow assets to the CMS. In my head, we should allow Links from SharedLinks to CMS Documents and those document to select an asset.

The problem is that we need more time to build that solution and time is a privilege we don't currently have so I hope you agree with this solution for the time being

JIRA: https://visitscotland.atlassian.net/browse/DS-964


jlrhurta (commented):
I would like to have an overview together about the solution as I have some questions around it

cms/src/main/java/com/visitscotland/brxm/validator/FileLinkValidator.java
```diff
@@ -0,0 +1,43 @@
...
+        }
+    }
+
+    private boolean hasValidLink (Node document, String fieldName) throws RepositoryException {
+        return document.hasNode(fieldName)
+                && document.getNode(fieldName).hasProperty("hippo:docbase")
+                && !document.getNode(fieldName).getProperty("hippo:docbase").getString().startsWith("cafebabe-");
+    }
+
+
```
martin-taylor-visitscotland, line 41: Just to highlight there's a tad too much white-space here (two lines)! 😄

cms/src/main/java/com/visitscotland/brxm/validator/FileLinkValidator.java
```diff
@@ -0,0 +1,43 @@
...
+                return Optional.of(context.createViolation());
+            } else {
+                return Optional.empty();
+            }
+        } catch (RepositoryException e) {
+            return Optional.of(context.createViolation());
+        }
+    }
+
+    private boolean hasValidLink (Node document, String fieldName) throws RepositoryException {
```
martin-taylor-visitscotland, line 35: A space is within the method signature:

```suggestion
    private boolean hasValidLink(Node document, String fieldName) throws RepositoryException {
```

cms/src/main/java/com/visitscotland/brxm/validator/FileLinkValidator.java
```diff
@@ -0,0 +1,43 @@
...
+            if (url && asset) {
+                return Optional.of(context.createViolation("both"));
+            } else if (!url && !asset) {
+                return Optional.of(context.createViolation());
+            } else {
+                return Optional.empty();
+            }
+        } catch (RepositoryException e) {
+            return Optional.of(context.createViolation());
+        }
```
martin-taylor-visitscotland, line 32: I think maybe a separate message should be used here other than the default to highlight an exceptional circumstance has taken place, thinking about it from a users perspective - they would just get a generic error message when it's not really true! I can also foresee this making things harder to debug should something go wrong!

site/components/src/main/java/com/visitscotland/brxm/hippobeans/FileLink.java
```diff
@@ -4,12 +4,22 @@
...
 import org.hippoecm.hst.content.beans.Node;
 import org.hippoecm.hst.content.beans.standard.HippoCompound;
+import org.hippoecm.hst.content.beans.standard.HippoBean;
 
 @HippoEssentialsGenerated(internalName = "visitscotland:FileLink")
 @Node(jcrType = "visitscotland:FileLink")
 public class FileLink extends HippoCompound implements UrlLink {
+
+    public static final String LINK = "visitscotland:link";
+    public static final String ASSET = "visitscotland:asset";
```
martin-taylor-visitscotland, line 14: I think we can have discussions about this at one of our technical discussions later 😄 however, I think these variables can be renamed to be clearer, such as `LINK_JCR_TYPE` or `LINK_JCR_PROPERTY_TYPE` - something along the lines of this, the more explicit the better!

What do you think?

site/components/src/main/java/com/visitscotland/brxm/services/LinkService.java
```diff
@@ -266,6 +271,16 @@ public String getPlainLink(Locale locale, HippoBean link, JsonNode product) {
         return processURL(locale, url);
     }
 
+    private String createLink(FileLink link) {
+        if (Contract.isEmpty(link.getLink())){
```
martin-taylor-visitscotland, line 275: I think we should consider gradually migrating away from `vs-utils` for functionality that is now natively supported in modern Java, such as using `Objects.isNull()` and `link.getLink().isBlank()`.

This would reduce our dependency on `vs-utils`, streamline the code-base, and ensure we leverage the latest JDK features. Given that `vs-utils` appears to be limited to Java SE 7, moving away from it would also position us better for future upgrades.

Thoughts? 😄

martin-taylor-visitscotland (approved):
All small comments that can be addressed at a later stage if you wish to merge.

jlrhurta (changes requested):
We are going to make CMS Assets a compound to make the CMS more intuitive as at the moment it is misleading

site/components/src/main/java/com/visitscotland/brxm/components/content/GeneralBSHContentComponent.java
```diff
@@ -45,10 +45,15 @@ public void doBeforeRender(HstRequest request, HstResponse response) {
...
-            HorizontalListLinksModule otyml = megalinkFactory.horizontalListLayout(page, request.getLocale());
-            request.setModel(OTYML_BUNDLE, otyml);
+        try {
+            GeneralBSH page = getDocument(request);
+            if (!Contract.isEmpty(page.getLinks())) {
+                HorizontalListLinksModule otyml = megalinkFactory.horizontalListLayout(page, request.getLocale());
+                request.setModel(OTYML_BUNDLE, otyml);
+            }
+        } catch (Exception e) {
+            //TODO remove try catch when troubleshooting is done
```
jcalcines, line 55: @jlrhurta , We probably want to link this comment to a ticket. I know there is no ticket at the moment so just keep this comment in the back of your head

site/components/src/main/java/com/visitscotland/brxm/hippobeans/FileLink.java
```diff
@@ -4,10 +4,12 @@
 import org.onehippo.cms7.essentials.dashboard.annotations.HippoEssentialsGenerated;
 import org.hippoecm.hst.content.beans.Node;
 import org.hippoecm.hst.content.beans.standard.HippoCompound;
+import org.hippoecm.hst.content.beans.standard.HippoBean;
```
jcalcines, line 7: This import can probably go away

cms/src/main/java/com/visitscotland/brxm/validator/FileLinkValidator.java
jcalcines: This validator has gone away, so no action needed here

jlrhurta (approved):

martin-taylor-visitscotland (approved):

PR #843, opened 2025-02-03
BSHUB-518 - Add price validator for events

martin-taylor-visitscotland:


jlrhurta (approved):

jlrhurta (approved):

cms/src/main/java/com/visitscotland/brxm/validator/EventPriceValidator.java
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * jcr:Name = visitscotland:event-price-validator
+ */
+public class EventPriceValidator implements Validator<Node> {
+    public EventPriceValidator() { }
+
+    @Override
+    public Optional<Violation> validate(ValidationContext context, Node node) {
+        try {
+            if(isPriceInvalid(node)) {
```
jcalcines, line 21: wee indentation issue. You are missin an space between if an the condition

cms/src/main/java/com/visitscotland/brxm/validator/EventPriceValidator.java
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * Checks if a price is valid
+     * @param node The JCR node to check
+     * @return true if the price is invalid, false otherwise
+     * @throws RepositoryException if there is an error accessing the JCR node
+     */
+    private boolean isPriceInvalid(final Node node) throws RepositoryException {
+        if(node.hasProperty("visitscotland:price")) {
+            final double price = node.getProperty("visitscotland:price").getDouble();
```
jcalcines, line 39: You probably could use a constant for the price

(Where that constant sits is a pending conversation)

cms/src/main/java/com/visitscotland/brxm/validator/EventPriceValidator.java
```diff
@@ -0,0 +1,45 @@
...
+            return Double.isNaN(price) || price < 0;
```
jcalcines, line 40: Can it be NaN and a primitive double at the same time?

cms/src/test/java/com/visitscotland/brxm/validator/EventPriceValidatorTest.java
```diff
@@ -0,0 +1,106 @@
...
+
+    private EventPriceValidator eventPriceValidator;
+
+    @BeforeEach
+    void setUp() {
+        this.eventPriceValidator = new EventPriceValidator();
+    }
+
+    @ValueSource(doubles = { POSITIVE_DOUBLE_VALUE, ZERO_DOUBLE_VALUE })
+    @ParameterizedTest
```
jcalcines, line 51: You are missing @DisplayName annotation

cms/src/main/java/com/visitscotland/brxm/validator/EventPriceValidator.java
```diff
@@ -0,0 +1,45 @@
...
+
+import javax.jcr.Node;
+import javax.jcr.RepositoryException;
+
+import java.util.Optional;
+
+/**
+ * jcr:Name = visitscotland:event-price-validator
+ */
+public class EventPriceValidator implements Validator<Node> {
```
jcalcines, line 15: Should this class be named just ´PriceValidator´? I know at the moment is only used for events, but It can have other applications in the future

jcalcines (approved):
@martin-taylor-visitscotland , I have added some comments but none of them need to be addressed or discussed as part of this PR. It is good to go

cms/src/main/java/com/visitscotland/brxm/validator/EventPriceValidator.java
martin-taylor-visitscotland: Great spot, let's think about implementing [CheckStyle](https://checkstyle.sourceforge.io/) at a later point to ensure this and standardise consistency within the code-base 😄!

One for the technical discussions!

cms/src/test/java/com/visitscotland/brxm/validator/EventPriceValidatorTest.java
martin-taylor-visitscotland: Happy to add it for consistency and as it is a standardised approach (that I keep forgetting) , but think we should note this for discussion at one of our technical discussions! What value does this bring over a clearly worded test name? 😄

cms/src/main/java/com/visitscotland/brxm/validator/EventPriceValidator.java
martin-taylor-visitscotland: Very valid point again, price is a compound type and can be reused - totally agree! Thanks for the points!

jcalcines (approved):

jcalcines (approved):

PR #850, opened 2025-02-05
DS-908 - Events Listings Module (Back-end)

jcalcines:


site/components/src/main/java/com/visitscotland/brxm/utils/PageTemplateBuilder.java
```diff
@@ -151,10 +155,13 @@ private void addModule(HstRequest request, PageConfiguration page, BaseDocument
             page.modules.add(devModuleFactory.getModule((DevModule) item));
         } else if (item instanceof CTABanner){
             page.modules.add(signPostFactory.createModule((CTABanner) item));
+        } else if (item instanceof EventsListing){
+            page.modules.add(eventsListingFactory.createModule((EventsListing) item));
```
DukesColleague, line 159: A side note not specifically relating to this PR - I think we should discuss this method in a technical discussion to figure out ways to split this up logically, the more modules come this method will exponentially grow, same with the factories!

site/components/src/main/java/com/visitscotland/brxm/services/ResourceBundleService.java
```diff
@@ -146,6 +151,10 @@ public String getResourceBundle(String bundleName, String key, Locale locale, bo
         return value;
     }
 
+//    private boolean ResourceBundleRegistry getRegistry(){
+//        return VsComponentManager.get(ResourceBundleRegistry.class);
+//    }
+
```
DukesColleague, line 157: Does this violate YAGNI - do we need this or can we remove this comment? If it's required, let's capture it so that it isn't forgotten about!

site/components/src/main/java/com/visitscotland/brxm/factory/event/EventsListingFactory.java
```diff
@@ -0,0 +1,216 @@
...
+        return filters;
+    }
+
+    private void addDateFields(List<EventFilter> filters) {
+        filters.add(buildDateField(START_DATE, START_DATE_LABEL));
+        filters.add(buildDateField(END_DATE, END_DATE_LABEL));
+    }
+
+
+    //TODO: Create a builder for this?
```
DukesColleague, line 160: I think this comment can be addressed, I agree with the builder pattern - In modern Java - I personally always favor fluent API as it massively helps with readability.

site/components/src/main/java/com/visitscotland/brxm/utils/PageTemplateBuilder.java
jcalcines: Good point! I have been thinking about this issue too. This method is now longer than "a screen" 😄 

The thing is this method is essentially a big switch-case block so it doesn't feel wrong either as it has a single purpose but I am happy to explore what are our options here at a later stage

site/components/src/main/java/com/visitscotland/brxm/utils/HippoUtilsService.java
```diff
@@ -44,19 +44,11 @@
  * Set of utilities related with Hippo that from the whole environment to be running in order to work
  */
 @Component
+//TODO: Move to services
```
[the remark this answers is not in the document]
jcalcines: Thank you for flagging. Probably not as part of this PR, as that change will result in a modification of many other files

https://github.com/visitscotland/dot-com-brxm/pull/855

site/components/src/main/java/com/visitscotland/brxm/services/ResourceBundleService.java
jcalcines: Good Spot! I was trying to solve an issue with resource bundles and I left that there by mistake

site/components/src/main/java/com/visitscotland/brxm/factory/event/EventsListingFactory.java
jcalcines: Yeah, I started pursuing that path but I got stucked. (and again we don't have much time) The current implementation is functional so I will classify this as technical debt?

Does that sound fair?

site/components/src/main/java/com/visitscotland/brxm/factory/event/EventsListingFactory.java
```diff
@@ -0,0 +1,216 @@
...
+public class EventsListingFactory {
+
+    // Value Lists
+    private final static String SECTORS_VALUE_LIST = "bsh-sectors";
+    private final static String TRAINING_TOPICS_VALUE_LIST = "bsh-training-topics";
+    private final static String INDUSTRY_EVENTS_TYPES_VALUE_LIST = "bsh-industry-event-types";
+    private final static String REGIONS_VALUE_LIST = "bsh-regions";
+
+    // Resource Bundle keys
+    private final static String BUNDLE = "events-listings";
```
jlrhurta, line 27: Should this bundle be bsh.events-listings?

jlrhurta (approved):

site/components/src/main/java/com/visitscotland/brxm/factory/event/EventsListingFactory.java
jcalcines: The prefix is automatically added from our ResourceBundleService

PR #852, opened 2025-02-06
(Version) Upgrade to version 15.7.1

jlrhurta:


site/components/src/main/java/com/visitscotland/brxm/hippobeans/BaseDocument.java
```diff
@@ -20,10 +20,17 @@ public class BaseDocument extends HippoDocument {
...
             return String.valueOf(node.getProperty("jcr:primaryType"));
         } catch (RepositoryException e) {
-            logger.error(String.format("jcr:primaryType has not been defined by the node %s ({%s})", getName(), getClass()));
+            try {
+                logger.error(String.format("jcr:primaryType has not been defined by the node %s ({%s})", getName(), getClass()));
+            } catch (Exception e2) {
+                logger.error("Error while getting primaryType", e2);
+            }
             return null;
         }
```
DukesColleague, line 35: I think it would be beneficial to simply be more verbose about what is actually being caught on line 25, for example:

```java
try { }
catch(RepositoryException exception) { logger.error(exception.getMessage, exception); }
```

site/components/src/main/java/com/visitscotland/brxm/hippobeans/GeneralBSH.java
```diff
@@ -41,9 +45,18 @@ public String[] getRegions() {
...
     @HippoEssentialsGenerated(internalName = "visitscotland:links", allowModifications = false)
     public List<HippoBean> getLinks() {
-        return getLinkedBeans("visitscotland:links", HippoBean.class);
+        //TODO: BSHUB-561 Temporal changes to troubleshoot the issue
+        Logger logger = LoggerFactory.getLogger("BSHUB-561");
+
+        try {
+            return getLinkedBeans("visitscotland:links", HippoBean.class);
+        } catch (Exception e) {
+            logger.error("Error while getting linked beans", e);
```
DukesColleague, line 54: `Exception` is too general, can we be more specific about what exceptions are going to be used? Also I think we can replace "Error while getting linked beans" with the exception message (or append) to add more value into clarifications!

If this is not here to stay can we also capture this within the [technical debt database](https://visitscotland.atlassian.net/wiki/spaces/VBE/database/472449052) so we can track? 😄

site/components/src/main/java/com/visitscotland/brxm/hippobeans/BaseDocument.java
jcalcines: Yeah, that's a fair point but we have to bear in mind that at the moment we are trying to troubleshoot an specific issue so I would go with the following:

`logger.error("Error while getting primaryType: " + e.getMessage, e);`
            
By doing this we make sure that we can search on the logs for an specific error message

site/components/src/main/java/com/visitscotland/brxm/hippobeans/GeneralBSH.java
jcalcines: Hey @DukesColleague, same as above, we need to capture an exception that we are not entirely confident what kind of exception is (It seems to be a NullPointerException)

Again, bear in mind that this code is not meant to stay. It will just help to troubleshoot the issue

DukesColleague: No worries, if it's not here to stay let's also document within [technical debt database](https://visitscotland.atlassian.net/wiki/spaces/VBE/database/472449052) so it's not lost in the abyss.

site/components/src/main/java/com/visitscotland/brxm/hippobeans/BaseDocument.java
```diff
@@ -20,10 +20,13 @@ public class BaseDocument extends HippoDocument {
     private final DocumentUtilsService documentUtils = VsComponentManager.get(DocumentUtilsService.class);
 
     public String getPrimaryType() {
+        //TODO: BSHUB-561 Temporal changes to troubleshoot the issue
+        Logger logger = LoggerFactory.getLogger("BSHUB-561");
+
         try {
             return String.valueOf(node.getProperty("jcr:primaryType"));
```
DukesColleague, line 27: Just a side note, one of the reasons I'm not a fan of calling methods in method parameter lists is because when something goes wrong, if say `var node = node.getProperty("jct:primaryType");` was on a separate line, I find it helps with debugging and reading traces!

DukesColleague (approved):

PR #855, opened 2025-02-07
Refactor HippoUtilsService

jcalcines:
This PR is just a tidy up work to move the HippoUtilsService in its functional package.

This class was created as a Singleton originally, but the purpose of this class is now wider, and it has been converted as a service so it can scale if needed.


site/components/src/main/java/com/visitscotland/brxm/services/HippoUtilsService.java
```diff
@@ -44,7 +43,6 @@
  * Set of utilities related with Hippo that from the whole environment to be running in order to work
  */
 @Component
```
DukesColleague, line 45: Can we use the `@Service` stereotype to be consistent with services?

DukesColleague (approved):

site/components/src/main/java/com/visitscotland/brxm/services/HippoUtilsService.java
jcalcines: @DukesColleague, There was something that prevented us from using Services in the past.

We need to validate that using `@Service` instead of `@Component`  would work before making such change

PR #863, opened 2025-02-15
BSHUB-561 - Site pages show 'Error'

jcalcines:
Hello Guys,

I have fixed several issues in this branch with the purpose of cleaning up the logs and to try to solve the problem.


repository-data/application/src/main/resources/hcm-actions.yaml
```diff
@@ -101,4 +101,7 @@ action-lists:
     /content/documents/administration/labels/global/pagination: reload
 - 2.5.5:
     /content/documents/administration/options/business-support-hub/sector: reload
+- 2.8.0:
+    /content/documents/administration/options/business-support-hub/global/global: reload
+    /content/documents/administration/bsh-labels/navigation/banner-dont-translate: delete
```
jlrhurta, line 106: Why are we removing the path to the banner? this is needed to make the banner work for bsh and we need a different one from .com as far as I understand it or we are going out without the banner?

site/components/src/main/java/com/visitscotland/brxm/services/LinkService.java
```diff
@@ -426,8 +426,11 @@ public Optional<EnhancedLink> createEnhancedLink(Linkable linkable, Module<?> mo
      * @param locale locale value
      * @return FlatLink simple format
      */
-    public FlatLink createSimpleLink(@NotNull Linkable linkable, Module<?> module, Locale locale) {
+    public FlatLink createSimpleLink(Linkable linkable, Module<?> module, Locale locale) {
```
DukesColleague, line 429: I think we should start adopting Jakarta EE annotations (similar to what was removed) for [`@Nullable`](https://jakarta.ee/specifications/annotations/3.0/annotations-spec-3.0#jakarta-annotation-nullable).

site/components/src/main/java/com/visitscotland/brxm/hippobeans/GeneralBSH.java
```diff
@@ -45,18 +45,9 @@ public String[] getRegions() {
...
-        Logger logger = LoggerFactory.getLogger("BSHUB-561");
-
-        try {
-            return getLinkedBeans("visitscotland:links", HippoBean.class);
-        } catch (Exception e) {
-            logger.error("Error while getting linked beans", e);
-            return Collections.emptyList();
-        }
+        return getLinkedBeans("visitscotland:links", HippoBean.class);
     }
```
DukesColleague, line 49: Not completely sure in the value of this method, can we use `getLinkedBeans("visitscotland:links", HippoBean.class)` where it needs to be called?

site/components/src/main/java/com/visitscotland/brxm/hippobeans/BaseDocument.java
```diff
@@ -1,33 +1,21 @@
...
-        try {
-            return String.valueOf(node.getProperty("jcr:primaryType"));
-        } catch (RepositoryException e) {
-            logger.error("Error while getting primaryType: {}", e.getMessage(), e);
-            return null;
-        }
-    }
-
-    protected static <T extends HippoBean> T getOnlyChild(List<T> children) {
+    protected <T extends HippoBean> T getOnlyChild(List<T> children) {
```
DukesColleague, line 18: This method signature is a bit confusing in my opinion -

* If the list of 'children' is empty, return `null` - fair enough! However, I would personally recommend `Optional<T>`.
* Else if the size of the children list is exactly 1, return the object at index 0 - fair enough.
* Else - the object at index `0`.

The third point seems as if it goes against what the signature is claiming to do - apparently it's the get the only child, however it is returning the first element of a potential list of many (side effect). I would recommend that the signature name is adjusted to reflect the behavior otherwise treat this as an exceptional circumstance and throw a clear custom exception.

site/components/src/main/java/com/visitscotland/brxm/components/content/GeneralBSHContentComponent.java
```diff
@@ -45,15 +45,10 @@ public void doBeforeRender(HstRequest request, HstResponse response) {
...
-            GeneralBSH page = getDocument(request);
-            if (!Contract.isEmpty(page.getLinks())) {
-                HorizontalListLinksModule otyml = megalinkFactory.horizontalListLayout(page, request.getLocale());
-                request.setModel(OTYML_BUNDLE, otyml);
-            }
-        } catch (Exception e) {
-            //TODO BSHUB-561 remove try catch when troubleshooting is done
-            logger.error("Error while generating Related pages (OTYML) for GeneralBSH", e);
+        GeneralBSH page = getDocument(request);
+        if (!Contract.isEmpty(page.getLinks())) {
```
DukesColleague, line 49: Just to highlight again from a previous discussion, I think we should slowly migrate away from using methods which are simple to do in core Java - this will eventually make our lives easier when it comes to maintaining - and thinking about what we do with the library this comes from which is using `Java SE 7` 🦖.

site/components/src/main/java/com/visitscotland/brxm/components/navigation/MenuComponent.java
```diff
@@ -87,12 +87,19 @@ private RootMenuItem getRootMenuItem(HstRequest request){
...
-        RootMenuItem rootMenuItem = factory.buildMenu(request, getResourceBundleID(request), id, cacheable);
+        RootMenuItem rootMenuItem = factory.buildMenu(request, getResourceBundleID(request), id, cacheable && !isSupportSite());
         rootMenuItem.setCmsCached(cacheable && editMode);
 
         return rootMenuItem;
     }
 
+    /**
+     * BSHUB-561
+     */
```
DukesColleague, line 98: This seems like a noise comment, unless there's a reason it's added - then the reason it's added should be added so that we know why it's caring about that ticket? 🤔

site/components/src/main/java/com/visitscotland/brxm/components/navigation/MenuComponent.java
```diff
@@ -87,12 +87,19 @@ private RootMenuItem getRootMenuItem(HstRequest request){
...
+    private boolean isSupportSite() {
+        return !Contract.isEmpty(siteProperties.getSiteId()) && siteProperties.getSiteId().equals("bsh");
```
DukesColleague, line 100: Highlighting my comment above, see comment regarding use of `Contract`.

site/components/src/main/java/com/visitscotland/brxm/services/DocumentUtilsService.java
```diff
@@ -28,9 +28,9 @@
 public class DocumentUtilsService {
     private static final Logger logger = LoggerFactory.getLogger(DocumentUtilsService.class.getName());
 
-    public static final String DOCUMENT_TYPE = "jcr:primaryType";
-    public static final String HIPPO_HANDLE = "hippo:handle";
-    public static final String HIPPO_FOLDER = "hippostd:folder";
+    static final String DOCUMENT_TYPE = "jcr:primaryType";
+    static final String HIPPO_HANDLE = "hippo:handle";
+    static final String HIPPO_FOLDER = "hippostd:folder";
```
DukesColleague, line 33: I think these should be private to promote loose coupling - what do you think? 😄

DukesColleague (commented):
Looks good 😄, I've put a few comments - more of a generic question, can we get some test coverage in please?

jlrhurta (approved):

DukesColleague (approved):

repository-data/application/src/main/resources/hcm-actions.yaml
jcalcines: The banner didn't have any effect on the project or was blocking the dot-com banner from working as they shared the same ID. Since it wasn't in use I though It would be better to remove it for the time being, and we can look into this at some other point?

Should we create a ticket for this?

site/components/src/main/java/com/visitscotland/brxm/hippobeans/BaseDocument.java
jcalcines: Hi @DukesColleague, I understand why you would think that's a better approach, but in this project we operate a policy of recovery over failing. 

The contract is that only one item should be provided from the field. However, the architecture from BloomReach is created so It can return several items and if we only need one we pick the first one. In case we detect that there are more than one element, we take the first one and we flag that the rest of them are going to be ignored. Bear in mind we don't even consider this to be a big issue as we are logging a warning instead of an error.

Throwing an exception when there is on more than one item would not provide any benefit and it would create an error on the CMS that upon detection we would need to troubleshoot and release a fix for it.

PR #869, opened 2025-02-20
BSHUB-627 - `EventDateValidator.java` forces the end-user to enter a date for an optional date.

DukesColleague:
It has been discovered that brXM defaults optional dates to 0001-01-01T12:00:00.000Z. The EventDateValidator.java retrieves the start and end dates of an event and checks whether the end date is after the start date. However, since the end date is an optional field in the CMS and defaults to 0001-01-01T12:00:00.000Z, this validation fails when a user enters an event without specifying an end date.

### DIFF
--- a/cms/src/main/java/com/visitscotland/brxm/validator/EventDateValidator.java
+++ b/cms/src/main/java/com/visitscotland/brxm/validator/EventDateValidator.java
@@ -15,6 +15,7 @@ import javax.jcr.Node;
 
 import java.util.Calendar;
 import java.util.Optional;
+import java.util.TimeZone;
 
 /**
  * jcrType = visitscotland:event-date-validator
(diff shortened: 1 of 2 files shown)


cms/src/main/java/com/visitscotland/brxm/validator/EventDateValidator.java
```diff
@@ -38,11 +39,22 @@ public Optional<Violation> validate(ValidationContext context, Node node) {
...
-    private boolean isEndDateAfterStartDate(final Node node) throws RepositoryException {
+    private boolean isEventDatesValid(final Node node) throws RepositoryException {
         final Calendar startDate = getStartDate(node);
         final Calendar endDate = getEndDate(node);
 
-        return endDate.after(startDate);
+        return isCalendarDefaultDate(endDate) || endDate.after(startDate);
+    }
+
+    private static boolean isCalendarDefaultDate(Calendar calendar) {
```
DukesColleague, line 49: Let's keep this final to keep it consistent!

```suggestion
    private static boolean isCalendarDefaultDate(final Calendar calendar) {
```

PR #871, opened 2025-02-21
BSHUB-612 - Events Listing - Create property for controlling the page size

jcalcines:
- Create property for controlling the page size
- Include hcm-actions to update labels related to this feature
- Amend the URLs to deliver fully-qualified URL for local environments
- BSH-566 - Update development data for Consents

### DIFF
--- a/site/components/src/main/java/com/visitscotland/brxm/rest/event/EventRepository.java
+++ b/site/components/src/main/java/com/visitscotland/brxm/rest/event/EventRepository.java
@@ -3,6 +3,7 @@ package com.visitscotland.brxm.rest.event;
 import com.visitscotland.brxm.hippobeans.EventBSH;
 import com.visitscotland.brxm.model.bsh.EventCard;
 import com.visitscotland.brxm.model.bsh.PaginatedResult;
+import com.visitscotland.brxm.utils.SiteProperties;
 import org.hippoecm.hst.component.support.bean.BaseHstComponent;
 import org.hippoecm.hst.content.beans.query.HstQueryResult;
 import org.hippoecm.hst.content.beans.standard.HippoBeanIterator;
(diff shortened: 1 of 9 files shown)


site/components/src/main/java/com/visitscotland/brxm/utils/SiteProperties.java
```diff
@@ -166,4 +166,9 @@ public String getDownloadExtensions() {
         return readString(DOWNLOAD_EXTENSIONS);
     }
 
+    public Integer getEventsListingsPageSize() {
```
DukesColleague, line 169: Is there a reason of why we are using the reference type for Integer over primitive? 

```suggestion
    public int getEventsListingsPageSize() {
```

site/components/src/main/java/com/visitscotland/brxm/rest/event/EventRepository.java
```diff
@@ -13,15 +14,16 @@
 @Component
 public class EventRepository extends BaseHstComponent {
 
-    //TODO: Create from configuration
```
DukesColleague, line 16: There's a wee TODO - just conscious that these aren't forgotten about 😄

site/components/src/main/java/com/visitscotland/brxm/utils/SiteProperties.java
```diff
@@ -166,4 +166,9 @@ public String getDownloadExtensions() {
...
+        int size = readInteger(EVENTS_LISTINGS_PAGE_SIZE);
+        return size > 0 ? size : 10;
```
jlrhurta, line 171: If the size is not defined in the CMS (size=8 expected) 10 items are expected? it that number right?

jlrhurta (approved):

site/components/src/main/java/com/visitscotland/brxm/utils/SiteProperties.java
DukesColleague: This is a good point, typically with pagination you can control the result number being sent back with a query parameter - would this not be easier to delegate this responsibility to the front-end?

jcalcines: @jlrhurta that value should never be used, it is just a placeholder. ReadInteger would return 0 if the property is not found, however returning 0 for a pagination could potentially cause an Arithmetic Exception. That's why I put a value. 

@DukesColleague this is just a fall-back value if none is configured, we could also delegate this on the front-end but when it is not defined we could make use of the default value retrieved from this property

jcalcines: @jlrhurta I won't make that change on this PR but I am working on reviewing outstanding work for Events-listings.

Do you still want me to change the placeholder to 8?

jlrhurta (approved):

duncansanderson (approved):

site/components/src/main/java/com/visitscotland/brxm/rest/event/EventRepository.java
jcalcines: That's my next task on https://visitscotland.atlassian.net/browse/BSHUB-614

PR #872, opened 2025-02-21
BSH-614 - Replace TODO with enhanced logging

jcalcines:


site/components/src/main/java/com/visitscotland/brxm/services/LinkService.java
```diff
@@ -580,15 +579,25 @@ private boolean hasOverrideImage(SharedLink sharedLink){
...
             module.addErrorMessage("This module should not contain a Video Link");
+            contentLogger.warn("This modules '{}' should not contain a Video Link", getSource(module));
         }
 
         return link;
     }
 
+    private String getSource(Module<?> module){
+        if (module == null || module.getHippoBean() == null){
+            //Note: if this value gets printed in the logs, we must refine the calculation of the source
```
DukesColleague, line 594: Not sure if this comment is worth having, when I read it I wasn't sure of it's intentions. Could we please remove this, or if it is required - add some context to clarify 😄

site/components/src/main/java/com/visitscotland/brxm/services/LinkService.java
```diff
@@ -580,15 +579,25 @@ private boolean hasOverrideImage(SharedLink sharedLink){
     private EnhancedLink enhancedLinkFromVideo(Video video, Module<?> module, Locale locale, boolean addCategory) {
         EnhancedLink link = createVideo(video, module, locale);
 
-        //TODO: Confirm requirements for Videos in HorizontalLinks VS-2086 indicates that no category is needed but we need to wait for the final designs before taking action.
         if (addCategory) {
+            //According to VS-2086 (old JIRA) no category is needed for videos
```
DukesColleague, line 583: Struggling to see the value in this comment, I don't think we should be referencing tickets in the code, what is the significance of having this? What does it tell us developers? Why should we be concerned with this ticket?

What do you think?

DukesColleague (approved):
A couple of small comments, about comments - very meta!

site/components/src/main/java/com/visitscotland/brxm/services/LinkService.java
jlrhurta: I think we should think about document the method if we feel it needs something to be explained?

jlrhurta (approved):

site/components/src/main/java/com/visitscotland/brxm/services/LinkService.java
jcalcines: In a nutshell, this is something that is not supposed to happen and it is preventing a NullPointerException when the module or the HippoBean are not pressent,

I am happy to take the comment out if you believe its obcuring things. Alternatively I could add a `log.warn` but It might be redundant as this calculation is done to log a message

site/components/src/main/java/com/visitscotland/brxm/services/LinkService.java
jcalcines: The comment is simply saying that "no category is needed for videos" and probably that's the best place to clarify that. 

"VS-2086 (old JIRA)" is just the source of the requirement for traceability purposes.

@DukesColleague, do you prefer me to refine or remove that comment?

DukesColleague: I would say that there's no need for the comment - we should be able to use git to pinpoint those changes paired with clear commit messages! It's good to keep the code as clear as possible IMO 😄 

That's just my two-cents though!

site/components/src/main/java/com/visitscotland/brxm/services/LinkService.java
DukesColleague: Sorry, what I meant was when reading this comment, I have no idea what "we must refine the calculation of the source" means 😄 

If we can add some context to clarify its intent I think that would be beneficial.

site/components/src/main/java/com/visitscotland/brxm/services/LinkService.java
jcalcines: I don't think I can improve the clarity of these messages without increasing the complexity of them, so I have remove both code commets

DukesColleague (approved):

PR #897, opened 2025-03-17
DS-1132 Solve issue with Calendar Dates and Time Zones

jcalcines:
It looks like the problem was related to the project using `Calendar` and `SimpleDateFormat` to format the dates.

Apparently, if we were using a more modern approach (`java.time`) we wouldn't have run into this issue. I don't think we can alter this as it is a platform feature but it would be nice to remember this for any other project

https://visitscotland.atlassian.net/browse/DS-1132

### DIFF
--- a/site/components/src/main/java/com/visitscotland/brxm/event/EventCardFactory.java
+++ b/site/components/src/main/java/com/visitscotland/brxm/event/EventCardFactory.java
@@ -12,18 +12,30 @@ import org.springframework.stereotype.Component;
 
 import java.text.SimpleDateFormat;
 import java.util.*;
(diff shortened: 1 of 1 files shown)


jcalcines:
This issue seems to be affecting only feature environments. I have come up with a fix but it might have a negative impact on cloud environments so it is on hold at the moment. 

I have created a [ticket with BloomReach](https://support.bloomreach.com/hc/en-us/requests/140889) to determine is there is a sensible way to overcome this situation, otherwise I might need to work with Webops to set up the timezone of the Environments to the same of the Cloud Environmets.

PR #898, opened 2025-03-17
Move valves to separate package to maintain clean package structure.

DukesColleague:
Small refactoring job to move `valves` into their own package to maintain a clean package structure - following the [boy scout rule](https://snappify.com/blog/boy-scout-rule).

### DIFF
--- a/site/components/src/main/java/com/visitscotland/brxm/valve/MountLocalizationValve.java
+++ b/site/components/src/main/java/com/visitscotland/brxm/valve/MountLocalizationValve.java
@@ -1,4 +1,4 @@
-package com.visitscotland.brxm.config;
+package com.visitscotland.brxm.valve;
 
 import org.apache.commons.lang3.LocaleUtils;
 import org.hippoecm.hst.container.valves.AbstractOrderableValve;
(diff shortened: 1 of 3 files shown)


jcalcines (approved):
@DukesColleague,

Why do you think these classes are better to be grouped in its own package? As far as I understand the purpose of this classes they are configuration of the Pipeline of the request to "intercept" request and enhance the information about the request.

Would it make sense to use a nested package such as `com.visitscotland.brxm.config.valve`?

Note: I am ok with merging this PR I was just wondering if an alternative way would be more consistent with this project's current setup

DukesColleague:
Hi @jcalcines - It's because I don't believe Servlet filters should be considered "configuration".

jcalcines:
@DukesColleague , Fair enough.

The only caveat is that the rest of the classes on the `config` package don't seem to match with the definition of configuration, as they seem to be project-wide tools used to alter the default general functionality of the framework. I guess following the same Boy Scout Principle, we can deal with those when the time comes. A potential solution to this might be to rename the package to something more meaningful

DukesColleague:
That's a great point @jcalcines, I totally agree with your observation - none of those classes should be considered configuration! That's the elegance of the boy scout rule, we make a couple small changes there and there and over time we'd hopefully see a huge difference and improvement in maintainability!

PR #900, opened 2025-03-18
BSHUB-718 - Clarify confusing VAT property on Events, added hint.

DukesColleague:


repository-data/application/src/main/resources/hcm-config/namespaces/visitscotland/EventBSH.yaml
```diff
@@ -298,9 +298,9 @@ definitions:
               size: '3'
           /vat:
             jcr:primaryType: frontend:plugin
-            caption: VAT Included?
+            caption: Price excludes VAT?
```
jcalcines, line 301: It is not natural language to exclude VAT. Prices can include or not VAT but they don't exclude it, therefore this field might be difficult to compute for content editors.

Also, In my head feel like a double negation and I struggle to think about when how to make the label to appear.

We probably should review this approach.

jcalcines (changes requested):

jlrhurta (approved):

jcalcines (approved):

PR #911, opened 2025-04-01
Prevent Error when branches have been merged to develop during the release

jcalcines:
I had the following error during the release

`[ERROR] Failed to execute goal com.amashchenko.maven.plugin:gitflow-maven-plugin:1.21.0:release-finish (default-cli) on project dot-com-brxm: release-finish: Remote branch 'origin/develop' is ahead of the local branch 'develop'. Execute git pull. -> [Help 1]`

This error occurs when branches are merged into develop between the executions of `start-release.sh` and `finish-release.sh`. 

While working on this, I also added an update for the master branch. Although this feature may not be technically useful for CI/CD pipelines, it could be helpful in cases where one developer starts the release and another finishes it.

**Note: In case this chage were accepted it would need to be propagated to other projects**
**Note 2: I had the same issue in dot-org-brxm. We might have to look at this differently but not with vs-dms-products**

## Summary by CodeRabbit

- **Chores**
  - Upgraded our automated release process to ensure that all critical components are synchronized before deployment.
  - Implemented enhanced error handling to promptly address issues during releases, resulting in smoother, more stable updates for end users.

### DIFF
--- a/ci/finish-release.bat
+++ b/ci/finish-release.bat
@@ -30,6 +30,32 @@ if errorlevel 1 (
   goto :EOF
 )
 
+:: Update main and develop branches
+echo echo "Updating main and develop branches"
+git checkout main
+if errorlevel 1 (
+  call :exit_on_failure "Checkout to main"
+  goto :EOF
+)
+
+git pull origin main
+if errorlevel 1 (
+  call :exit_on_failure "Pulling main"
+  goto :EOF
+)
+
+git checkout develop
+if errorlevel 1 (
+  call :exit_on_failure "Checkout to develop"
+  goto :EOF
+)
+
+git pull origin develop
+if errorlevel 1 (
+  call :exit_on_failure "Pulling develop"
+  goto :EOF
+)
+
 :: Checkout to the release branch
 echo Checking out the release branch %releaseBranch%
 git checkout "%releaseBranch%"
--- a/ci/finish-release.bat
+++ b/ci/finish-release.bat
@@ -31,7 +31,7 @@ if errorlevel 1 (
 )
 
 :: Update main and develop branches
-echo echo "Updating main and develop branches"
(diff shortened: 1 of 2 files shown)


AuldCode:
My initial thoughts are that when the local develop branch is out of sync with its counterpart on the remote (with the remote being ahead), how can we be certain that the release branch has develop's latest additions? With the proposed update (never mind the .bat version for now, as I am actively working on the batch start and finish release versions), the main branch and the develop branch will indeed be in sync with the remote, but am I in the wrong here to not feel so secure about the release branch having those latest updates? 

As @jcalcines says, if one person starts the release, then another pulls the release branch, in-between the first person commits changes (may or may not bring them over to the release branch as well), then it's not just that the develop and main branches in the second person's system are out of sync with the remote, the release branch might be outdated too. This is the case in this execution, as there was no conflict when the release branch was committed to the remote, but a git pull in the release branch might be necessary as well.

As this is a file that is being copied over from one repo to another, it is essential that is fixed for all repos. I will need to look into this new way of the script being executed and would advise against any immediate action taken. Let me know though, if such a change needs to happen straight away.

AuldCode (commented):
My thoughts here. Leaving the PR open for now, until we address the issues raised across other repositories that use the same approach.

https://github.com/visitscotland/dot-com-brxm/pull/911#issuecomment-2769766608

jcalcines:
> My thoughts here. Leaving the PR open for now, until we address the issues raised across other repositories that use the same approach.
> 
> [#911 (comment)](https://github.com/visitscotland/dot-com-brxm/pull/911#issuecomment-2769766608)

@AuldCode I have converted this PR to a draft to prevent an accidental merge

jcalcines:
@gavin-vs  I am closing this PR since the last merge to develop is performing this operation.

This branch is no longer needed

PR #916, opened 2025-04-02
DS-1185 - Enhance content fallback logic focusing on separation of concerns testability.

DukesColleague:


site/components/src/main/java/com/visitscotland/brxm/translation/UnableToMatchMountException.java
```diff
@@ -0,0 +1,13 @@
+package com.visitscotland.brxm.translation;
+
+import org.hippoecm.hst.core.request.ResolvedVirtualHost;
+
+final class UnableToMatchMountException extends TranslationException {
+    private static final String EXCEPTION_MESSAGE =
+        "failed to resolve request path '%s' to a mount that is on the virtual host %s";
```
DukesColleague, line 7: What if the Virtual Host is null?

site/components/src/main/java/com/visitscotland/brxm/translation/TranslationFallbackProvider.java
```diff
@@ -0,0 +1,87 @@
...
+            .map(Mount::getContentPath)
+            .orElseThrow(() -> new UnableToMatchMountException(requestPath, virtualHost));
+    }
+
+    private String getRelativeContentPath(final HstRequestContext requestContext) {
+        return Optional
+            .ofNullable(requestContext.getResolvedSiteMapItem())
+            .map(ResolvedSiteMapItem::getRelativeContentPath)
+            .orElseThrow(RelativeContentPathNotFoundException::new);
+    }
```
DukesColleague, line 76: These methods are probably useful to have extracted for reuse elsewhere, although until then I'm following YAGNI and avoiding premature optimization until such a time that they are required.

PR #918, opened 2025-04-04
Integrate Assets

jcalcines:
## Summary by CodeRabbit

- **New Features**
  - Introduced enhanced asset linking that now displays expanded file details, including file size and type.
  - Enabled improved external document size calculations with human-readable outputs.
  - Added components that ensure more reliable URL generation and link management.

- **Refactor**
  - Refined error handling for shared links and unavailable external documents to provide clearer, consistent messaging to end-users.


site/components/src/main/java/com/visitscotland/brxm/utils/NonTestable.java
```diff
@@ -29,7 +29,11 @@ enum Cause {
          * The purpose of the method is to wrap other service so the code is not tightly couple to other services
          * (i.e HippoCms, DMS, Instagram)
          */
-        BRIDGE;
+        BRIDGE,
+        /**
+         * The purpose of the method is to wrap a static method to increase testability.
+         */
+        STATIC;
```
DukesColleague, line 36: Let's add this to the agenda for our talks on testing!

site/components/src/main/java/com/visitscotland/brxm/services/LinkService.java
```diff
@@ -613,15 +622,20 @@ public String getDownloadText(String link) {
...
+                module.addErrorMessage(errorMessage);
             }
-            contentLogger.warn("The external document {} might be broken.", link);
+            contentLogger.warn(errorMessage);
             return "";
         } else {
-            return " | " + size;
+            return " | " + sizeType.get();
         }
     }
```
DukesColleague, line 640: Probably would be good for this method to return an `Optional<String>` instead of returning an empty String.

Also I'd recommend we stay away from formatting on the back-end, referring to the pipe `|` here.

site/components/src/main/java/com/visitscotland/brxm/services/HttpConnectionProvider.java
```diff
@@ -0,0 +1,20 @@
...
+@Component
+@NonTestable(NonTestable.Cause.STATIC)
+public class HttpConnectionProvider {
+
+    public HttpURLConnection openConnection(String url) throws IOException {
+        HttpURLConnection connection = (HttpURLConnection) new URL(url).openConnection();
+        // This Timeout (5s) prevents the CMS from freezing if the connection is unstable
+        connection.setConnectTimeout(5000);
+        return connection;
+    }
```
DukesColleague, line 19: Are you sure this is `NonTestable`? I can think of a few valuable tests I think! 😄 

Personally, I would have the connection timeout be a configurable property rather than hard-coded, providing flexibility if adjustment is required.

site/components/src/main/java/com/visitscotland/brxm/services/FileSizeCalculator.java
```diff
@@ -0,0 +1,79 @@
...
+        DecimalFormatSymbols dfs = new DecimalFormatSymbols(locale);
+        DecimalFormat decimalFormat = new DecimalFormat("#.#", dfs);
+
+        if (bytes > 0) {
+            String size = decimalFormat.format(bytes / 1024 / 1024) + "MB";
+            return type.map(s -> s + " " + size).or(() -> Optional.of(size));
+        }
+
+        return Optional.empty();
+    }
```
DukesColleague, line 78: I recommend formatting is handled by it's consumers - however I can see the value in converting the value to MB if required. I would say a clearer method name could be provided rather than `composeText()` perhaps `calculateSizeInMegaBytes()`?

site/components/src/main/java/com/visitscotland/brxm/services/FileSizeCalculator.java
```diff
@@ -0,0 +1,79 @@
...
+    private final HttpConnectionProvider connectionProvider;
+
+    public FileSizeCalculator(HttpConnectionProvider connectionProvider) {
+        this.connectionProvider = connectionProvider;
+    }
+
+    /**
+     * Calculates the Size of the External document if the document
+     *
+     */
```
DukesColleague, line 30: No need for this comment - the method name is descriptive enough!

`getExternalDocumentSize()`.

site/components/src/main/java/com/visitscotland/brxm/services/CommonUtilsService.java
```diff
@@ -95,29 +95,29 @@ public String buildQueryString(Map<String, String> parameters, String encoding)
...
                 displayType = link.substring(link.lastIndexOf(".")+1).toUpperCase();
             }
-            return displayType + " " + decimalFormat.format(bytes / 1024 / 1024) + "MB";
+            return Optional.of(displayType + decimalFormat.format(bytes / 1024 / 1024) + "MB");
         } catch (IOException e) {
             logger.error("The URL {} is not valid", link, e);
         }
-        return null;
+        return Optional.empty();
     }
```
DukesColleague, line 121: Possible side-note or refactoring job - this method is humongous! Possible violation of doing one and only one thing 😄

site/components/src/main/java/com/visitscotland/brxm/services/AssetLinkFactory.java
```diff
@@ -0,0 +1,67 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+import javax.jcr.RepositoryException;
+import java.util.Locale;
+import java.util.Optional;
+
+@Component
+public class AssetLinkFactory {
```
DukesColleague, line 16: Perhaps for another technical discussion, we may need to rethink of design patterns here - I think it's not great that these factories are being created exponentially whenever a new module is required - could be an indicator of a bit of a code smell.

site/components/src/main/java/com/visitscotland/brxm/utils/VSRequestContextProvider.java
```diff
@@ -0,0 +1,26 @@
...
+
+import javax.jcr.Node;
+
+@Component
+@NonTestable(NonTestable.Cause.BRIDGE)
+public class VSRequestContextProvider {
+
+    public HstRequestContext getRequestContext() {
+        return RequestContextProvider.get();
+    }
```
DukesColleague, line 16: Is this method really required? What value does it add opposed to simply calling `RequestContextProvider.get()` 😄

site/components/src/main/java/com/visitscotland/brxm/utils/VSRequestContextProvider.java
```diff
@@ -0,0 +1,26 @@
...
+
+    public HstLink getHstLink(Node node){
+        return getRequestContext().getHstLinkCreator().create(node, getRequestContext());
+    }
+
+    public String getUrl (Node node){
```
DukesColleague, line 22: Small space here!

```suggestion
    public String getUrl(Node node){
```

Also, I think there can be valuable tests added this!

site/components/src/main/java/com/visitscotland/brxm/services/CommonUtilsService.java
jcalcines: Spot on! FileSizeCalculator will replace this class

site/components/src/main/java/com/visitscotland/brxm/services/FileSizeCalculator.java
jcalcines: aha! I need to rename the class to `FileMetaDataCalculator`

The class not only get the size but also the *File type*

site/components/src/main/java/com/visitscotland/brxm/services/HttpConnectionProvider.java
jcalcines: I have created a follow-up tricket fro unit test
https://visitscotland.atlassian.net/browse/BSHUB-774

site/components/src/main/java/com/visitscotland/brxm/services/FileMetaDataCalculator.java
```diff
@@ -0,0 +1,90 @@
...
+    @Cacheable(value="externalDocument")
+    public Optional<String> getDisplayText(String link, Locale locale, boolean includeType) {
+        FileMetaData metaData = getFileMetaData(link);
+        if (metaData != null) {
+            Optional<String> displayType = includeType? getType(metaData.getMimeType(), link) : Optional.empty();
+            return composeText(displayType, metaData.getContentLength(), locale);
+        }
+
+        return Optional.empty();
+    }
```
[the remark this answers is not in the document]
jcalcines: Out of the Scope for this piece of work

site/components/src/main/java/com/visitscotland/brxm/services/FileMetaDataCalculator.java
```diff
@@ -0,0 +1,90 @@
...
+    private Optional<String> getType(String mimeType, String link){
+        if (!mimeType.startsWith("application") && !mimeType.startsWith("image")) {
+            return Optional.empty();
+        } else if (mimeType.contains("pdf")) {
+            return Optional.of("PDF");
+        } else if (link.contains(".")) {
+            return Optional.of(link.substring(link.lastIndexOf(".")+1).toUpperCase());
+        }
+        return Optional.empty();
+    }
```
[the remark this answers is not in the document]
jcalcines: Using a mimetype is a non-requested added feature. The actual request was to use the file extension (which is not reliable enough).

This is the compromise we reached

site/components/src/main/java/com/visitscotland/brxm/services/FileMetaDataCalculator.java
```diff
@@ -0,0 +1,90 @@
...
+            return new FileMetaData(link, connection.getContentLength(), connection.getContentType());
+        } catch (IOException e) {
+            logger.error("The URL {} is not valid", link, e);
+        } finally {
+            if (connection != null) {
+                connection.disconnect();
+            }
+        }
+        return null;
+    }
```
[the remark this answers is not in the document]
jcalcines: I have some auxiliar classes already setting those Timeouts and the current error handling is fulfilling the requirements

site/components/src/main/java/com/visitscotland/brxm/services/FileMetaDataCalculator.java
```diff
@@ -0,0 +1,90 @@
...
+        final DecimalFormatSymbols decimalFormatSymbols = new DecimalFormatSymbols(locale);
+        final DecimalFormat decimalFormat = new DecimalFormat("#.#", decimalFormatSymbols);
+
+        if (bytes > 0) {
+            String size = decimalFormat.format(bytes / BYTES_PER_MB) + "MB";
+            return fileType.map(type -> type + " " + size).or(() -> Optional.of(size));
+        }
+
+        return Optional.empty();
+    }
```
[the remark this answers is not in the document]
jcalcines: Agree with the idea but the requirement is to show always the size in MB

site/components/src/main/java/com/visitscotland/brxm/services/FileSizeCalculator.java
jcalcines: Done!

site/components/src/main/java/com/visitscotland/brxm/utils/VSRequestContextProvider.java
jcalcines: We have agreed on the use of this.(For the time being)

site/components/src/main/java/com/visitscotland/brxm/model/bsh/FileMetaData.java
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    public long getContentLength() {
+        return contentLength;
+    }
+
+    public String getMimeType() {
+        return mimeType;
+    }
+}
```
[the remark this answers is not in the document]
jcalcines: This class will potentially become a record. I will add a TODO instead

site/components/src/main/java/com/visitscotland/brxm/services/AssetLinkFactory.java
```diff
@@ -0,0 +1,77 @@
...
+import com.visitscotland.brxm.utils.VsRequestContextProvider;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+import javax.jcr.RepositoryException;
+import java.util.Locale;
+
+@Component
+public class AssetLinkFactory {
```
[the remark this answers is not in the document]
jcalcines: Yeah, that's in the roadmap for review, at the moment it is better to keep consistency until we define the way forward

site/components/src/main/java/com/visitscotland/brxm/services/AssetLinkFactory.java
```diff
@@ -0,0 +1,76 @@
...
+        if (sharedLink != null && sharedLink.getLinkType() instanceof Asset){
+            return (Asset)sharedLink.getLinkType();
+        } else {
+            throw new UnsupportedOperationException("Link type not supported");
+        }
+    }
+
+    private String getLabel(String text, AssetLink link, Locale locale) {
+        return fileSizeCalculator.getDisplayText(link, locale).map(sizeType -> text + " | " + sizeType).orElse(text);
+    }
```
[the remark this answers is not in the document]
jcalcines: This is based on an old solution and it is bound to be retired. This solution is easier to adapt/remove when the time comes

site/components/src/main/java/com/visitscotland/brxm/services/AssetLinkFactory.java
```diff
@@ -0,0 +1,66 @@
...
+
+    private String getMimeType(Asset asset) {
+        try {
+            return asset.getAsset().getNode().getNode("hippogallery:asset").getProperty("jcr:mimeType").getString();
+        } catch (RepositoryException | NullPointerException e) {
+            log.warn("The mimetype for the asset {} is not available: Error: {}", asset.getPath(), e.getMessage());
+        }
+        return null;
+    }
```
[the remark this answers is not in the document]
jcalcines: Haha, we are review our approach to Optional Framework and creating that method as optional doesn't bring any benefic

site/components/src/main/java/com/visitscotland/brxm/services/AssetLinkFactory.java
```diff
@@ -0,0 +1,66 @@
...
+            return asset.getAsset().getNode().getNode("hippogallery:asset").getProperty("jcr:data").getLength();
+        } catch (RepositoryException | NullPointerException e) {
+            log.warn("The size of the asset {} could not be calculated. Error: {}", asset.getPath(), e.getMessage());
+        }
+        return 0L;
+    }
+
+    private String getURL(Asset asset){
+        return requestContextProvider.getUrl(asset.getAsset().getNode());
+    }
```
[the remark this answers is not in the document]
jcalcines: You are Shifting the NPE from this method to the caller and this is actually the origin of the problem.

site/components/src/main/java/com/visitscotland/brxm/services/CommonUtilsService.java
```diff
@@ -101,30 +104,31 @@ public String getExternalDocumentSize(String link, Locale locale) {
...
             if (!contentType.startsWith("application") && !contentType.startsWith("image")) {
-                return null;
+                return Optional.empty();
             } else if (concatFormat && contentType.contains("pdf")) {
                 displayType = "PDF " ;
             } else if (concatFormat && link.contains(".")) {
                 displayType = link.substring(link.lastIndexOf(".")+1).toUpperCase() + " ";
             }
-            return displayType + decimalFormat.format(bytes / 1024 / 1024) + "MB";
+            return Optional.of(displayType + decimalFormat.format(bytes / 1024 / 1024) + "MB");
```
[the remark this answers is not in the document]
jcalcines: True, but this is a deprecated method. Its replacement would improve this logic

site/components/src/main/java/com/visitscotland/brxm/services/LinkService.java
```diff
@@ -519,11 +521,16 @@ private EnhancedLink enhancedLinkFromPage(Page page, Module<?> module, Locale lo
...
     private EnhancedLink enhancedLinkFromSharedLink(SharedLink sharedLink, Module<?> module, Locale locale, boolean addCategory) {
-        EnhancedLink link = new EnhancedLink();
+        EnhancedLink link;
         JsonNode product = getNodeFromSharedLink(sharedLink, module, locale);
 
-        link.setTeaser(sharedLink.getTeaser());
-        link.setLink(getPlainLink(locale, sharedLink.getLinkType(), product));
+        if (sharedLink.getLinkType() instanceof Asset){
+            link = assetLinkFactory.create(sharedLink, locale);
+        } else {
```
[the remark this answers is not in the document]
jcalcines: No need to do so. the factory will always return an object

repository-data/development/src/main/resources/hcm-content/content/documents/bsh/sandbox.yaml
```diff
@@ -5460,16 +5466,17 @@
...
             visitscotland:regions: []
             visitscotland:sectors: []
             visitscotland:skill: ''
             visitscotland:source: ''
-            visitscotland:teaser: Dummy Document
+            visitscotland:teaser: This is a empty word document! Don't look at it!
+              It is just empty. Vacio.
             visitscotland:title: Dummy Word Document
             visitscotland:topic: []
             visitscotland:type: ''
```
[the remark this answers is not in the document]
jcalcines: It is intentional

repository-data/development/src/main/resources/hcm-content/content/documents/bsh/sandbox.yaml
```diff
@@ -5392,15 +5396,16 @@
...
```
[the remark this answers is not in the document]
jcalcines: It is intentional

site/components/src/main/java/com/visitscotland/brxm/factory/ArticleFactory.java
```diff
@@ -187,39 +189,68 @@ private void validate(ArticleModule module){
             }
         }
     }
-    
+
+    private String getPublishDate(SharedLink sharedLink, Locale locale) {
```
DukesColleague, line 193: I think a good signature for this would be:

`public Optional<String> getPublishDate(final SharedLink sharedLink,
								   final Locale locale)`

site/components/src/test/java/com/visitscotland/brxm/services/CommonUtilsServiceTest.java
```diff
@@ -83,13 +83,13 @@ void getExternalDocumentSize(){
         when(huc.getContentType()).thenReturn(MediaType.APPLICATION_PDF_VALUE);
 
         when(huc.getContentLength()).thenReturn(1024*1024*3);
```
DukesColleague, line 85: Let's extract this magic number out into a test constant!

site/components/src/test/java/com/visitscotland/brxm/services/CommonUtilsServiceTest.java
```diff
@@ -83,13 +83,13 @@ void getExternalDocumentSize(){
...
-        Assertions.assertEquals("PDF 3MB", utils.getExternalDocumentSize("pdf", Locale.UK));
+        Assertions.assertEquals("PDF 3MB", utils.getExternalDocumentSize("pdf", Locale.UK).get());
 
         when(huc.getContentLength()).thenReturn(1024*512);
-        Assertions.assertEquals("PDF 0.5MB", utils.getExternalDocumentSize("pdf", Locale.UK));
+        Assertions.assertEquals("PDF 0.5MB", utils.getExternalDocumentSize("pdf", Locale.UK).get());
 
         when(huc.getContentLength()).thenReturn(1024*512*201);
```
DukesColleague, line 91: Possibly out-of-scope, but what is `huc`? 😆

site/components/src/test/java/com/visitscotland/brxm/services/CommonUtilsServiceTest.java
```diff
@@ -99,25 +99,25 @@ void getExternalDocumentSize_decimalRounding(){
 
         //800Kb File
```
DukesColleague, line 100: Static test constant!

site/components/src/test/java/com/visitscotland/brxm/services/CommonUtilsServiceTest.java
```diff
@@ -99,25 +99,25 @@ void getExternalDocumentSize_decimalRounding(){
...
         when(huc.getContentLength()).thenReturn(1024*800);
-        Assertions.assertEquals("PDF 0.8MB", utils.getExternalDocumentSize("/file.pdf", Locale.UK));
+        Assertions.assertEquals("PDF 0.8MB", utils.getExternalDocumentSize("/file.pdf", Locale.UK).get());
 
         //Almost 10 mb
```
DukesColleague, line 104: Static test constant!

site/components/src/main/java/com/visitscotland/brxm/utils/VsRequestContextProvider.java
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    public HstLink getHstLink(Node node){
+        if (node != null) {
+            return getRequestContext().getHstLinkCreator().create(node, getRequestContext());
+        }
+        return null;
+    }
+
+    public String getUrl(Node node){
```
DukesColleague, line 25: Can we make this method name a little clearer in terms of what it means from `VsRequestContextProvider`?

`VsRequestContextProvider.getUrl(Node node)`.

My first initial question would be, why are we getting a URL from a node in the context of `VsRequestContextProvider`?

site/components/src/main/java/com/visitscotland/brxm/utils/VsRequestContextProvider.java
```diff
@@ -0,0 +1,33 @@
...
+
+@Component
+@NonTestable(NonTestable.Cause.BRIDGE)
+public class VsRequestContextProvider {
+
+    public HstRequestContext getRequestContext() {
+        return RequestContextProvider.get();
+    }
+
+    public HstLink getHstLink(Node node){
```
DukesColleague, line 18:
```suggestion
    public Optional<HstLink> getHstLink(Node node){
```

site/components/src/main/java/com/visitscotland/brxm/utils/VsRequestContextProvider.java
```diff
@@ -0,0 +1,33 @@
...
+
+import javax.jcr.Node;
+
+@Component
+@NonTestable(NonTestable.Cause.BRIDGE)
+public class VsRequestContextProvider {
+
+    public HstRequestContext getRequestContext() {
+        return RequestContextProvider.get();
+    }
```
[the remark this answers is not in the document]
jcalcines: The purpose of this method is to encapsulate this abstract class. There is no need of further verification

site/components/src/main/java/com/visitscotland/brxm/services/LinkService.java
```diff
@@ -620,15 +627,28 @@ public String getDownloadText(String link) {
...
             if (module != null) {
-                module.addErrorMessage("The Link to the External document might be broken");
+                module.addErrorMessage(errorMessage);
             }
-            contentLogger.warn("The external document {} might be broken.", link);
+            contentLogger.warn(errorMessage);
             return "";
         } else {
-            return " | " + size;
+            return " | " + sizeType.get();
```
DukesColleague, line 651: Not a huge fan of this, we shouldn't be doing formatting on our side!

site/components/src/main/java/com/visitscotland/brxm/factory/ArticleFactory.java
jcalcines: From the back of our conversation yesteday, I'm not too keen to return an Optional as the value returned by this method is used to populate a DTO and it doesn't seem add value to convert this to an optional.

I'm also not too keen on adding `final` to the parameters , but I know it is a common practice so happy to discuss :)

site/components/src/test/java/com/visitscotland/brxm/services/CommonUtilsServiceTest.java
jcalcines: I don't think this needs a Constant as it is a value only used for this unit test and it is easy to interpret as 3MB. What to you think?

site/components/src/test/java/com/visitscotland/brxm/services/CommonUtilsServiceTest.java
jcalcines: That's fair! but, might the comment be simpler to understand than the constant.

      final int SIZE_800_KB = 1024*800;
      when(huc.getContentLength()).thenReturn(SIZE_800_KB);
      Assertions.assertEquals("PDF 0.8MB", utils.getExternalDocumentSize("/file.pdf", Locale.UK).get());

site/components/src/main/java/com/visitscotland/brxm/services/HttpConnectionProvider.java
```diff
@@ -0,0 +1,42 @@
...
+import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.net.URL;
+
+@Component
+@NonTestable(NonTestable.Cause.WRAP)
+public class HttpConnectionProvider {
+
+    private static final int DEFAULT_CONNECTION_TIMEOUT_MS = 5000;
+    private static final int DEFAULT_READ_TIMEOUT_MS = 10000;
```
DukesColleague, line 19: I would argue these should be configurable, aka. properties / environment variables.

site/components/src/test/java/com/visitscotland/brxm/services/CommonUtilsServiceTest.java
jcalcines: That's fair! but, might the comment be simpler to understand than the constant?

      final int ALMOST_3MB = 1024*1024*10 - 30;
      when(huc.getContentLength()).thenReturn(ALMOST_3MB);
      Assertions.assertEquals("PDF 0.8MB", utils.getExternalDocumentSize("/file.pdf", Locale.UK).get());

site/components/src/main/java/com/visitscotland/brxm/services/HttpConnectionProvider.java
```diff
@@ -0,0 +1,42 @@
...
+
+    private final int connectionTimeoutMs;
+    private final int readTimeoutMs;
+
+    public HttpConnectionProvider() {
+        this(DEFAULT_CONNECTION_TIMEOUT_MS, DEFAULT_READ_TIMEOUT_MS);
+    }
+
+    public HttpConnectionProvider(int connectionTimeoutMs, int readTimeoutMs) {
```
DukesColleague, line 28: Could be more flexible to take a `Duration` instead of an `int` 😄

site/components/src/test/java/com/visitscotland/brxm/services/CommonUtilsServiceTest.java
```diff
@@ -99,25 +99,25 @@ void getExternalDocumentSize_decimalRounding(){
...
         when(huc.getContentLength()).thenReturn(1024*1024*10 - 30);
-        Assertions.assertEquals("PDF 10MB", utils.getExternalDocumentSize("/file.pdf", Locale.UK));
+        Assertions.assertEquals("PDF 10MB", utils.getExternalDocumentSize("/file.pdf", Locale.UK).get());
 
         //Slightly less than 10 mb
```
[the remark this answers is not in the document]
jcalcines: That's fair! but, might the comment be simpler to understand than the constant?

  final int BIGGER_THAN_3MB = 1024*1024*10 + 30;
  when(huc.getContentLength()).thenReturn(BIGGER_THAN_3MB);
  Assertions.assertEquals("PDF 3.0MB", utils.getExternalDocumentSize("/file.pdf", Locale.UK).get());

site/components/src/main/java/com/visitscotland/brxm/utils/VsRequestContextProvider.java
jcalcines: What about `getUrlFromNode`?

site/components/src/main/java/com/visitscotland/brxm/services/FileMetaDataCalculator.java
```diff
@@ -0,0 +1,101 @@
...
+    public FileMetaDataCalculator(HttpConnectionProvider connectionProvider) {
+        this.connectionProvider = connectionProvider;
+    }
+
+    private FileMetaData getFileMetaData(String link) {
+        HttpURLConnection connection = null;
+        try {
+            connection = connectionProvider.openConnection(link);
+            if (connection.getResponseCode() >= 400) {
+                return null;
```
DukesColleague, line 33: I would say that this is an exceptional circumstance, should we throw an exception?

site/components/src/main/java/com/visitscotland/brxm/services/FileMetaDataCalculator.java
```diff
@@ -0,0 +1,101 @@
...
+            }
+            return new FileMetaData(link, connection.getContentLength(), connection.getContentType());
+        } catch (IOException e) {
+            logger.error("The URL {} is not valid", link, e);
+        } finally {
+            if (connection != null) {
+                connection.disconnect();
```
DukesColleague, line 40: Can't we use `try-with-resources`? Just thinking out loud!

site/components/src/main/java/com/visitscotland/brxm/utils/VsRequestContextProvider.java
jcalcines: I understand why you would think this could be optional, but the only way of getting a null is to pass a null value as parameter. 

Do you think it is valuable to convert the method to Optional?

site/components/src/main/java/com/visitscotland/brxm/utils/NonTestable.java
```diff
@@ -29,7 +29,11 @@ enum Cause {
          * The purpose of the method is to wrap other service so the code is not tightly couple to other services
          * (i.e HippoCms, DMS, Instagram)
          */
-        BRIDGE;
+        BRIDGE,
+        /**
+         * The purpose of the method is to wrap a static method or a class instance to increase testability.
+         */
+        WRAP;
```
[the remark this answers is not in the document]
jcalcines: Yeah, already captured (I don't know why your previous comment dissappeared

site/components/src/main/java/com/visitscotland/brxm/services/LinkService.java
jcalcines: I know what you mean I am not a big fan either but we have some production code already relying on this composition.

The work I have done on this PR is to prepare ourselves to move away from this, but we need to apply the same principle to download links and also to communicate this to the front-end.

However, there is a port of me that thinks that this is necessary. It happens that this composition is also used when we detect that an external document has been linked from a copy (RichTextEditor) and we enhance the title of the link adding the filetype and the size to meet some accessibility requirement. 

If we were to fully move away from this practice, we need to work with Front End developers to design a solution for this.

site/components/src/main/java/com/visitscotland/brxm/services/HttpConnectionProvider.java
jcalcines: Yeah, I'm just trying not to keep growing the ball of changes

site/components/src/main/java/com/visitscotland/brxm/services/HttpConnectionProvider.java
jcalcines: I will look into the difference, but integer sounds simpler?

site/components/src/main/java/com/visitscotland/brxm/services/FileMetaDataCalculator.java
jcalcines: I tried but It turn out that `HttpConnection` doesn't implement the auto closeable feature

site/components/src/main/java/com/visitscotland/brxm/services/FileMetaDataCalculator.java
jcalcines: Our policy in this proyect is to recover from errors instead of propagating them, so we better stick with that?

I know we could get roughly the same by throwing an Exception but my concern is that the potential uses of this class are managing their way with the expected output `null` and we might need to do a big refactoring if we go down this route

DukesColleague: * Clarity and Explicitness: Exceptions clearly signal that something unexpected has occurred, making it easier for developers to identify and handle critical issues. In contrast, error recovery might mask problems, leading to silent failures that are harder to diagnose.
* Separation of Concerns: Throwing exceptions allows the error-handling logic to be separated from the main code flow. This can make the code cleaner and more maintainable, as error handling is not mixed with the primary functionality.
* Flexibility in Handling: Exceptions can be caught and handled at different levels of the application, giving developers flexibility in how they manage errors. This flexibility is particularly useful in complex systems where different components might need to handle errors differently.
* Preventing Data Corruption: In some cases, continuing to execute code after an error can lead to data corruption or inconsistent states. Throwing an exception ensures that the application stops executing potentially harmful code, preserving data integrity.
* Documentation and Debugging: Exceptions often provide detailed information about what went wrong, which can be invaluable for debugging. This can lead to faster resolution of issues and better documentation of error conditions.
* User Experience: While error recovery might seem user-friendly, it can sometimes lead to confusing behavior if not properly communicated. Exceptions, when handled correctly, can provide clear feedback to users about what went wrong, helping them understand and potentially resolve the issue.

site/components/src/main/java/com/visitscotland/brxm/services/HttpConnectionProvider.java
DukesColleague: Depends on the flexibility, this way it's always milliseconds. It's a side point though, happy for you to dismiss 😄

site/components/src/main/java/com/visitscotland/brxm/services/HttpConnectionProvider.java
DukesColleague: Please take it as a side-note that we can come back to

site/components/src/main/java/com/visitscotland/brxm/services/LinkService.java
DukesColleague: Let's get a meeting in with the UI devs!

site/components/src/main/java/com/visitscotland/brxm/services/FileMetaDataCalculator.java
jcalcines: Even though I agree with these principles, I think still get some benefits from proper error recovery. Usually the principle in this project is to try to deliver as much value as possible to the user so, when we cannot provide some information we just ignore it because we can still deliver some valuable information.

In my head, The exception propagation and encapsulation works well when you try to perform an operation that cannot be completed for some reason. I that situation you want to feed back as much information as possible to simplify troubleshooting and that's critical.

However, this application is even more complex than a Website, as we have several types of users who would benefit from different kind of information:
- Final users: They don't care about development problems, they just want to get the data. Error pages just put them off as the site would look unstable
- CMS users: They benefit from content related feedback such as "This document you are linking to doesn't seem to be available, and we are hiding it from you". These users benefit from *preview error messages* (WIP?) and the Content Logger (WIP)
- Developers: They benefit from information about code doing stuff that is not expected. They shouldn't care about an external document not being available at a point in time.

Obviously the user that suffers the most about not having proper error propagation are the developers (closely followed by CMS Users) so let's try to pay attention to what we log

Please, don't get me wrong, I still agree with having a proper exception management, but this project might not be ideal for extensive error propagation.

jcalcines: As I was writing this, I was analysing how an exception would help in this case and even though the propagation wouldn't go far I kind of see how we can benefit from it in this case.

DukesColleague (approved):
As previously discussed on multiple occasions, I have expressed concerns about shipping untested code and making assumptions about its functionality. In this instance, I am approving the request to allow the requester to proceed at their own discretion. However, I want to emphasise that this approval is contingent on the understanding that the requester acknowledges and accepts the potential risks associated with deploying untested code. It is the requester's responsibility to ensure thorough testing and validation before implementation.

site/components/src/test/java/com/visitscotland/brxm/services/CommonUtilsServiceTest.java
DukesColleague: That works too! I'd say, if >1 test uses this as a scenario, let's make it a test constant - if not, let's keep it local :)

jcalcines:
> As previously discussed on multiple occasions, I have expressed concerns about shipping untested code and making assumptions about its functionality. In this instance, I am approving the request to allow the requester to proceed at their own discretion. However, I want to emphasise that this approval is contingent on the understanding that the requester acknowledges and accepts the potential risks associated with deploying untested code. It is the requester's responsibility to ensure thorough testing and validation before implementation.

@DukesColleague , Thanks for flagging it. This ticket has passed QA and it ready to be merged from their point of view.

In relation to the Unit Tests missing on this ticket I have created a follow-up ticket for it https://visitscotland.atlassian.net/browse/BSHUB-774

PR #926, opened 2025-04-10
DS-1218 - Add JAX-RS resource for exposing resource bundles.

DukesColleague:


site/components/src/main/java/com/visitscotland/brxm/rest/ResourceBundleEndpoint.java
```diff
@@ -0,0 +1,35 @@
...
+    protected ResourceBundleEndpoint(ResourceBundleService resourceBundleService) {
+        this.resourceBundleService = resourceBundleService;
+    }
+
+    @GET
+    @Path("/{bundleId}")
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response getBundle(final @PathParam("bundleId") String bundleId,
+                              final @QueryParam("locale") String locale) {
+        final var parsedLocale = Locale.forLanguageTag(locale);
```
DukesColleague, line 30: What happens if `locale` is null? How do we handle that scenario?

site/components/src/main/java/com/visitscotland/brxm/rest/ResourceBundleEndpoint.java
```diff
@@ -0,0 +1,35 @@
...
+        final var result = resourceBundleService.getAllLabels(bundleId, parsedLocale);
```
DukesColleague, line 31: No resource bundle? Let's be sensible and return a `404`.

site/components/src/main/java/com/visitscotland/brxm/rest/ResourceBundleEndpoint.java
```diff
@@ -0,0 +1,35 @@
...
+    private final ResourceBundleService resourceBundleService;
+
+    protected ResourceBundleEndpoint(ResourceBundleService resourceBundleService) {
+        this.resourceBundleService = resourceBundleService;
+    }
+
+    @GET
+    @Path("/{bundleId}")
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response getBundle(final @PathParam("bundleId") String bundleId,
```
DukesColleague, line 28: `Null`? Let's return a `400` with some detail on what went wrong.

PR #932, opened 2025-04-14
BSHUB-611 - Page error: CTA link to unpublished page

jcalcines:
I converted the Signpost Module into an Error Module so we can give context to CMS Users and we log the issue as a Content Issue

JIRA: https://visitscotland.atlassian.net/browse/BSHUB-611

### DIFF
--- a/site/components/src/main/java/com/visitscotland/brxm/factory/SignpostFactory.java
+++ b/site/components/src/main/java/com/visitscotland/brxm/factory/SignpostFactory.java
@@ -5,14 +5,18 @@ import com.visitscotland.brxm.hippobeans.capabilities.Linkable;
 import com.visitscotland.brxm.model.FlatImage;
 import com.visitscotland.brxm.model.FlatLink;
 import com.visitscotland.brxm.model.LinkType;
+import com.visitscotland.brxm.model.Module;
 import com.visitscotland.brxm.model.SignpostModule;
+import com.visitscotland.brxm.model.ErrorModule;
 import com.visitscotland.brxm.utils.AnchorFormatter;
 import com.visitscotland.brxm.services.LinkService;
 import com.visitscotland.brxm.services.ResourceBundleService;
+import com.visitscotland.brxm.utils.ContentLogger;
 import com.visitscotland.brxm.utils.HippoHtmlWrapper;
 import com.visitscotland.brxm.services.HippoUtilsService;
 import com.visitscotland.brxm.utils.SiteProperties;
 import com.visitscotland.utils.Contract;
+import org.slf4j.Logger;
 import org.springframework.stereotype.Component;
 
 import java.util.Locale;
@@ -22,6 +26,7 @@ public class SignpostFactory {
 
     private static final String BUNDLE_ID = "newsletter-signpost";
 
+    private final Logger contentLogger;
     private final ResourceBundleService bundle;
     private final SiteProperties properties;
     private final HippoUtilsService hippoUtilsService;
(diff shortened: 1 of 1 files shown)


jlrhurta (approved):

site/components/src/main/java/com/visitscotland/brxm/factory/SignpostFactory.java
```diff
@@ -62,16 +69,20 @@ public SignpostModule createDeliveryAPIModule(Locale locale) {
         return createSignPostModule(properties.getSiteId() +"." + BUNDLE_ID, "newsletter", locale);
     }
 
-    public SignpostModule createModule (CTABanner ctaBanner){
+    public Module<?> createModule (CTABanner ctaBanner){
         SignpostModule module = new SignpostModule();
         Linkable linkable = (Linkable) ctaBanner.getCtaLink().getLink();
         FlatLink cta = linkService.createSimpleLink(linkable, module, null);
-        if (!Contract.isEmpty(ctaBanner.getCtaLink().getLabel())) {
```
DukesColleague, line 69: Side note, I think we should move away from this ideology of a custom `Contract` object in favor of what could be done just as easily using native language features - helping us decouple the dependency on this 'Contract'!

site/components/src/main/java/com/visitscotland/brxm/factory/SignpostFactory.java
```diff
@@ -62,16 +69,20 @@ public SignpostModule createDeliveryAPIModule(Locale locale) {
...
-            cta.setLabel(ctaBanner.getCtaLink().getLabel());
-        }
 
         if (Contract.isNull(cta.getLink())) {
-            return null;
+            contentLogger.warn(String.format(
+                    "The link for the CTA banner %s is not available. The module has been hidden", ctaBanner.getPath()));
+            return new ErrorModule(ctaBanner,
```
DukesColleague, line 80: What is the `ErrorModule` used for again?

DukesColleague (approved):
As previously discussed on multiple occasions, I have expressed concerns about shipping untested code and making assumptions about its functionality. In this instance, I am approving the request to allow the requester to proceed at their own discretion. However, I want to emphasize that this approval is contingent on the understanding that the requester acknowledges and accepts the potential risks associated with deploying untested code. It is the requester's responsibility to ensure thorough testing and validation before implementation.

site/components/src/main/java/com/visitscotland/brxm/factory/SignpostFactory.java
jcalcines: Yeah, that's a good suggestion but the native features are not quite the same.

I think the sensible approach at this point is try to transition to the Optional Framework

site/components/src/main/java/com/visitscotland/brxm/factory/SignpostFactory.java
jcalcines: `ErrorModule` functionality is twofold:
- Final Users Indicates to the front-end that a module should go in that place but it should be ignored.
- CMS Users: A message would appear in the CMS flagging the issue to Content Editors. This funcionality is only currently available for the Freemarker implementation but we are expecting to bring this to the headless approach

jcalcines:
> As previously discussed on multiple occasions, I have expressed concerns about shipping untested code and making assumptions about its functionality. In this instance, I am approving the request to allow the requester to proceed at their own discretion. However, I want to emphasize that this approval is contingent on the understanding that the requester acknowledges and accepts the potential risks associated with deploying untested code. It is the requester's responsibility to ensure thorough testing and validation before implementation.

A ticket has been create to reflect the Missing Unit Test https://visitscotland.atlassian.net/browse/BSHUB-782

site/components/src/main/java/com/visitscotland/brxm/factory/SignpostFactory.java
```diff
@@ -22,6 +26,7 @@ public class SignpostFactory {
 
     private static final String BUNDLE_ID = "newsletter-signpost";
 
+    private final Logger contentLogger;
```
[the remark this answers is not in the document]
jcalcines: This log has a different purpose from the general logger. Please don't flag again this issue:
 "Consider naming consistency for the logger"

DukesColleague (approved):

PR #936, opened 2025-04-15
DS-1226 Improve `ResourceBundleService` to enhance clarity, testability, readability and maintainability

DukesColleague:


site/components/src/main/java/com/visitscotland/brxm/resource/ResourceBundleQueryService.java
```diff
@@ -0,0 +1,158 @@
...
+    Map<String, String> getAllValuesFor(final @Nonnull String bundleName, final Locale locale, final boolean isSiteBundle) {
+        final String resolvedBundleName = resolveBundleName(bundleName, isSiteBundle);
+
+        try {
+            final ResourceBundle resourceBundle = resourceBundleRegistry.getBundle(resolvedBundleName, locale);
+            return resourceBundleValueExtractor.extractValuesFromResourceBundleAsMap(resourceBundle);
+        } catch (final MissingResourceException | ClassCastException exception) {
+            throw new ResourceQueryFailedException(bundleName, exception);
+        }
+    }
```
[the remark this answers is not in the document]
DukesColleague: Very nice spot, I'll add `@Nonnull` to the Locale.

site/components/src/test/java/com/visitscotland/brxm/resource/ResourceBundleQueryServiceTest.java
```diff
@@ -0,0 +1,257 @@
...
+    @Test
+    void getAllValuesFor_ThrowsResourceQueryFailedException_WhenExtractValuesFromResourceBundleAsMapThrowsClassCastException() {
+        final var resourceBundle = Mockito.mock(ResourceBundle.class);
+
+        when(registry.getBundle(BUNDLE_NAME)).thenReturn(resourceBundle);
+        when(valueExtractor.extractValuesFromResourceBundleAsMap(resourceBundle))
+            .thenThrow(ClassCastException.class);
+
+        Assertions.assertThrowsExactly(ResourceQueryFailedException.class, () ->
+            resourceBundleQueryService.getAllValuesFor(BUNDLE_NAME, IS_NOT_SITE_BUNDLE));
```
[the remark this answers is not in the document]
DukesColleague: What on earth are you talking about?

site/components/src/main/java/com/visitscotland/brxm/resource/ResourceBundleQueryService.java
```diff
@@ -0,0 +1,158 @@
...
+     * @return a map containing all keys and their corresponding string values from the resolved resource bundle for the specified locale
+     * @throws ResourceQueryFailedException if the resource bundle cannot be found or if the contents
+     * cannot be cast to strings, wrapping the original exception
+     */
+    Map<String, String> getAllValuesFor(final @Nonnull String bundleName, final Locale locale, final boolean isSiteBundle) {
+        final String resolvedBundleName = resolveBundleName(bundleName, isSiteBundle);
+
+        try {
+            final ResourceBundle resourceBundle = resourceBundleRegistry.getBundle(resolvedBundleName, locale);
+            return resourceBundleValueExtractor.extractValuesFromResourceBundleAsMap(resourceBundle);
```
jcalcines, line 149: Same as `getValueFor` however for this one you need to be extra clever as some of the values might exist in the language but not some others

site/components/src/main/java/com/visitscotland/brxm/resource/ResourceBundleQueryService.java
```diff
@@ -0,0 +1,158 @@
...
+
+import java.util.MissingResourceException;
+import java.util.ResourceBundle;
+import java.util.Optional;
+import java.util.Locale;
+import java.util.Map;
+
+@Component
+class ResourceBundleQueryService {
+    private final ResourceBundleValueExtractor resourceBundleValueExtractor;
```
jcalcines, line 17: The class is missing a feature when a label might be optional and would only be used when defined.

But we have talk about that requirement and I am happy for it to be implemented when you come across to a case when that is needed. We might get another workaround to that feature,  now that we are moving to the resource API

site/components/src/main/java/com/visitscotland/brxm/resource/ResourceException.java
```diff
@@ -0,0 +1,7 @@
+package com.visitscotland.brxm.resource;
+
+public abstract class ResourceException extends RuntimeException {
```
jcalcines, line 10: This exception doesn't seem to be used?

site/components/src/main/java/com/visitscotland/brxm/resource/ResourceBundleValueExtractor.java
```diff
@@ -0,0 +1,70 @@
...
+     */
+    Map<String, String> extractValuesFromResourceBundleAsMap(final @Nonnull ResourceBundle resourceBundle) {
+        final Map<String, String> bundleMap = new HashMap<>();
+
+        for (final String key : resourceBundle.keySet()) {
+            bundleMap.put(key, resourceBundle.getString(key));
+        }
+
+        return bundleMap;
+    }
```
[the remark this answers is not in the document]
DukesColleague: In this occasion I am choosing NOT to proceed with this suggestion, I think streams and functional approaches are great where it enhances readability, unfortunately in this case I think it does not.

I will proceed with the traditional imperative solution.

site/components/src/test/java/com/visitscotland/brxm/resource/ResourceBundleQueryServiceTest.java
```diff
@@ -0,0 +1,243 @@
...
+    void getValueFor_ValidBundleNameItemKeyAndNotSiteBundle_OptionalOfValue() {
+        final var resourceBundle = Mockito.mock(ResourceBundle.class);
+        final var expected = "See more...";
+
+        when(registry.getBundle(BUNDLE_NAME)).thenReturn(resourceBundle);
+        when(valueExtractor.extractValueFromResourceBundle(resourceBundle, ITEM_KEY)).thenReturn(Optional.of(expected));
+
+        var actual = resourceBundleQueryService.getValueFor(BUNDLE_NAME, ITEM_KEY, false);
+
+        Assertions.assertAll(
```
jcalcines, line 55: Do you need this block? `AssertAll` is the default option. Should we keep this for complex assertions?

site/components/src/test/java/com/visitscotland/brxm/resource/ResourceBundleQueryServiceTest.java
```diff
@@ -0,0 +1,243 @@
...
+            () -> Assertions.assertTrue(actual.isPresent()),
+            () -> Assertions.assertEquals(expected, actual.orElseThrow())
+        );
+
+        verify(registry).getBundle(eq(BUNDLE_NAME));
```
jcalcines, line 60: I don't think this brings any value. Line 50 automatically ensures that this method is called at least once

jcalcines: Note: There are several instances of this issue I won't comment in all of them

site/components/src/test/java/com/visitscotland/brxm/resource/ResourceBundleQueryServiceTest.java
jcalcines: Note: There are several instances of this issue I won't comment in all of them

site/components/src/test/java/com/visitscotland/brxm/resource/ResourceBundleQueryServiceTest.java
```diff
@@ -0,0 +1,243 @@
...
+        verify(valueExtractor).extractValueFromResourceBundle(eq(resourceBundle), eq(ITEM_KEY));
+    }
+
+    @Test
+    void getValueFor_ValidBundleNameItemKeyLocaleAndNotSiteBundle_OptionalOfValue() {
```
jcalcines, line 65: This is a side not for further discussion, but the method name might be a bit hard to read. (It took me a bit to get that the difference is the "Locale" word.

We could benefit here for more Human readable descriptions with @DisplayName

site/components/src/test/java/com/visitscotland/brxm/resource/ResourceBundleQueryServiceTest.java
```diff
@@ -0,0 +1,243 @@
...
+            () -> Assertions.assertTrue(actual.isPresent()),
+            () -> Assertions.assertEquals(expected, actual.orElseThrow())
+        );
+
+        verify(siteProperties).getSiteId();
+        verify(registry).getBundle(eq(SITE_BUNDLE_NAME));
+    }
+
+    @ParameterizedTest
+    @MethodSource("nullParameterCombinationsForGetValueForWithLocale")
```
jcalcines, line 104: Can you define `nullParameterCombinationsForGetValueForWithLocale` just below this method?

site/components/src/test/java/com/visitscotland/brxm/resource/ResourceBundleQueryServiceTest.java
```diff
@@ -0,0 +1,243 @@
...
+
+        final var actual = resourceBundleQueryService.getAllValuesFor(BUNDLE_NAME, UNITED_KINGDOM, IS_SITE_BUNDLE);
+
+        Assertions.assertEquals(expected, actual);
+        verify(siteProperties).getSiteId();
+        verify(registry).getBundle(eq(SITE_BUNDLE_NAME), eq(UNITED_KINGDOM));
+    }
+
+    @Test
+    void getAllValuesFor_ReturnsBundleValuesAsMap_IfValidBundleNameAndLocaleAndSiteBundleFlagIsFalse() {
```
jcalcines, line 177: Should you use When instead of If in `IfValidBundleNameAndLocaleAndSiteBundleFlagIsFalse`?

site/components/src/test/java/com/visitscotland/brxm/resource/ResourceBundleValueExtractorTest.java
```diff
@@ -0,0 +1,184 @@
...
+
+    private ResourceBundleValueExtractor resourceBundleValueExtractor;
+
+    @BeforeEach
+    void setUp() {
+        this.resourceBundleValueExtractor = new ResourceBundleValueExtractor();
+    }
+
+    @Test
+    void When_ExtractValueFromResourceBundle_With_ValidResourceBundleAndUnknownKey_OptionalOfEmpty() {
```
jcalcines, line 35: I have noticed that the name convention for this file is different from the previous one, Is that intended?

site/components/src/test/java/com/visitscotland/brxm/resource/ResourceBundleValueExtractorTest.java
```diff
@@ -0,0 +1,184 @@
...
+class ResourceBundleValueExtractorTest {
+    private static final String UNKNOWN_KEY = "unknown";
+    private static final String KNOWN_KEY = "known";
+
+    @Mock private ResourceBundle resourceBundle;
+
+    private ResourceBundleValueExtractor resourceBundleValueExtractor;
+
+    @BeforeEach
+    void setUp() {
```
jcalcines, line 30: Why do you need this? can you use `@InjectMocks` instead of creating this method?

site/components/src/test/java/com/visitscotland/brxm/resource/ResourceBundleQueryServiceTest.java
```diff
@@ -0,0 +1,243 @@
...
+        final var expected = "See more...";
+
+        when(registry.getBundle(BUNDLE_NAME)).thenReturn(resourceBundle);
+        when(valueExtractor.extractValueFromResourceBundle(resourceBundle, ITEM_KEY)).thenReturn(Optional.of(expected));
+
+        var actual = resourceBundleQueryService.getValueFor(BUNDLE_NAME, ITEM_KEY, false);
+
+        Assertions.assertAll(
+            () -> Assertions.assertTrue(actual.isPresent()),
+            () -> Assertions.assertEquals(expected, actual.orElseThrow())
```
jcalcines, line 57: Why don't you use just `actual.get()`?

jcalcines (approved):

site/components/src/test/java/com/visitscotland/brxm/resource/ResourceBundleQueryServiceTest.java
DukesColleague: It will complain that you cannot use `.get()` without checking `.isPresent()` check, but I figured instead of suppressing it'll be good if it just throws and let's me know that way!

site/components/src/test/java/com/visitscotland/brxm/resource/ResourceBundleValueExtractorTest.java
DukesColleague: Not really, as this class has no dependents it can't 'inject' anything! So I just want to instantiate a new instance of the service and use the `resourceBundle` as a global mock!

site/components/src/test/java/com/visitscotland/brxm/resource/ResourceBundleValueExtractorTest.java
DukesColleague: Great spot, I changed half way as I didn't like the clarity of them, let me change up for consistency!

site/components/src/test/java/com/visitscotland/brxm/resource/ResourceBundleQueryServiceTest.java
DukesColleague: Great spot again @jcalcines - I'd say it should typically be `when` other than when it makes sense to use `if`.

site/components/src/main/java/com/visitscotland/brxm/resource/ResourceBundleQueryService.java
```diff
@@ -0,0 +1,158 @@
...
+     * @param isSiteBundle a flag indicating whether the bundle is site-specific, which affects
+     * how the bundle name is resolved
+     * @return an {@code Optional} containing the value associated with {@code itemKey} if found;
+     * otherwise, an empty {@code Optional}
+     * @throws ResourceQueryFailedException if the resource bundle cannot be found or if the contents
+     * cannot be cast to strings, wrapping the original exception
+     */
+    Optional<String> getValueFor(final String bundleName, final String itemKey, final boolean isSiteBundle) {
+        if (Contract.anyNull(bundleName, itemKey)) {
+            return Optional.empty();
```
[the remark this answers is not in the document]
DukesColleague: Yes, you are correct! I'm going to edit this to throw an NPE!

site/components/src/main/java/com/visitscotland/brxm/resource/ResourceBundleQueryService.java
```diff
@@ -0,0 +1,158 @@
...
+import javax.annotation.Nonnull;
+
+import java.util.MissingResourceException;
+import java.util.ResourceBundle;
+import java.util.Optional;
+import java.util.Locale;
+import java.util.Map;
+
+@Component
+class ResourceBundleQueryService {
```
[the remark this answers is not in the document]
DukesColleague: Yes, it should be public! I'm start creating classes using the strictest access modifiers and elevate when required, so one I start using this in the other classes and they need visibility I'll elevate 😄

site/components/src/test/java/com/visitscotland/brxm/resource/ResourceBundleQueryServiceTest.java
DukesColleague: Another great point, perhaps this is a good use case, let me evaluate!

site/components/src/main/java/com/visitscotland/brxm/resource/ResourceException.java
DukesColleague: It's the base exception for all resource-based exceptions, it is abstract and extended through `ResourceQueryFailedException`.

site/components/src/main/java/com/visitscotland/brxm/resource/ResourceBundleQueryService.java
```diff
@@ -0,0 +1,158 @@
...
+                                 final boolean isSiteBundle) {
+        if (Contract.anyNull(bundleName, itemKey, locale)) {
+            return Optional.empty();
+        }
+
+        final String resolvedBundleName = resolveBundleName(bundleName, isSiteBundle);
+
+        try{
+            final ResourceBundle resourceBundle = resourceBundleRegistry.getBundle(resolvedBundleName, locale);
+            return resourceBundleValueExtractor.extractValueFromResourceBundle(resourceBundle, itemKey);
```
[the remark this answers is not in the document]
DukesColleague: Hi, I think this case would be captured here. If a Resource Bundle does not exist it will throw a MissingResourceException, that will then be wrapped in a ResourceQueryFailedException and should be handled by the caller.

site/components/src/main/java/com/visitscotland/brxm/resource/ResourceBundleQueryService.java
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    private String resolveBundleName(final @Nonnull String bundleName, final boolean isSiteBundle) {
+        return isSiteBundle ? getSiteBundleName(bundleName) : bundleName;
+    }
+
+    private String getSiteBundleName(final @Nonnull String bundleName) {
+        return siteProperties.getSiteId() + "." + bundleName;
+    }
+}
```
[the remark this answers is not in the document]
DukesColleague: I think this implementation is sensible, it does exactly that.

Take `resourceBundleValueExtractor.extractValueFromResourceBundle(resourceBundle, itemKey)`, for example, it returns an `Optional<String>`, if the specified `key` is not present within the given `resourceBundle` then `Optional.empty()` is returned.

In the case that a `resourceBundle` cannot be extracted from the registry, I think this should be considered an exceptional circumstance.

With that said, I do not agree with your suggestion - the main reason being if a `MissingResourceException` or `ClassCastException` is thrown, in your example, it'll simply return `Optional.empty()`. This is dangerous. Why would we choose to silently fail? Think about the engineer utilising this method. It can make debugging a nightmare.

jlrhurta (approved):

site/components/src/test/java/com/visitscotland/brxm/resource/ResourceBundleQueryServiceTest.java
```diff
@@ -0,0 +1,246 @@
...
+import java.util.stream.Stream;
+import java.util.Optional;
+import java.util.Objects;
+import java.util.Locale;
+import java.util.Map;
+
+import static org.mockito.Mockito.when;
+
+@ExtendWith(MockitoExtension.class)
+class ResourceBundleQueryServiceTest {
```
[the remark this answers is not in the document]
DukesColleague: No.

jlrhurta (approved):

PR #960, opened 2025-06-17
feat(hero): if a static video is present, use the new hero section

mjpvs:
A requirement has been raised to use the new hero section on the home page of .com, as there will now always be a video present which should autoplay and shouldn't be reliant on cookie acceptance like the existing video component. This adjustment switches to the new component if a video is present which doesn't come from youtube (along with the back end 

Juan Luis and I have raised some concerns about creating technical debt here by ending up in a situation where we're using both the old and new headers long term, and that transition should be on the the radar moving forward. I've also requested a dev team catchup once James is back so we can pin down our process for this kind of transition moving forward, and to touch back on our previous discussion about how we should track technical debt.

### DIFF
--- a/repository-data/application/src/main/resources/hcm-content/content/documents/visitscotland/shared.yaml
+++ b/repository-data/application/src/main/resources/hcm-content/content/documents/visitscotland/shared.yaml
@@ -135,6 +135,85 @@
       hippostd:foldertype: [new-translated-folder-videos, new-video]
       hippotranslation:id: 31f821a8-1fb4-4667-8b06-8e4856f9c3ea
       hippotranslation:locale: en
+      /orkney:
+        jcr:primaryType: hippo:handle
+        jcr:mixinTypes: ['hippo:named', 'hippo:versionInfo', 'mix:referenceable']
+        jcr:uuid: 367a959d-985e-463d-94d1-13bb603ab044
+        hippo:name: Orkney
+        hippo:versionHistory: 27c7f8ff-9dc2-4a07-a86d-b701a50c287b
+        /orkney[1]:
+          jcr:primaryType: visitscotland:Video
+          jcr:mixinTypes: ['mix:referenceable']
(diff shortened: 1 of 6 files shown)


markup-mitchell:
@mjpvs apart from technical debt, we should very cautious about autoplaying video as it's widely regarded as undesirable. 

IMO this sort of decision should go through a formal review process involving named representatives of UX and UI. If a decision is made to go against those recommendations someone should be prepared to formally responsible for that choice. 

https://www.a11yproject.com/posts/never-use-auto-play/

@jameshaddrillvs

jlrhurta (approved):

GemmaLouise:
@mjpvs The Page Intro component will eventually be deprecated in favour of the new Hero Section - I'd like to make sure this is clear in our roadmap/versioning so that the product owners are aware of work needed to make the switch. 

@markup-mitchell  The auto-playing video has already gone through a UX testing process and lots of research has gone into this around accessibility and the testing will continue in a live environment with real users as this is an iterative process. The video hero follows WCAG guidelines which requires us to provide a pause button for auto-playing videos. It should only ever be used on the homepage, see documentation here https://visitscotland.atlassian.net/wiki/spaces/DE/pages/224821254/Hero+section#With-video

GemmaLouise (approved):

mjpvs:
Thanks @GemmaLouise , from our conversation I believe that they are aware of that plan but not necessarily of how we'll actually go about that deprecation or what might need to be changed in the existing content to make the new one work. It sounds like that will all be covered in your handover so that's all good, especially if this is only ever going to be used on the home page then we probably aren't introducing too much extra unnecessary complexity here with this check for different kinds of videos. It was just tricky to answer some questions about requirements for that transition, while explaining why it needs to happen to avoid keeping the two versions live long term, without the inside track or being sure what Michael knew about the new component which was making us a little nervous here

PR #1003, opened 2025-09-09
Add import statement

markup-mitchell:
This defect wasn't noticed because in all previous test environments there was already another example of a megalinks layout on the page.


markup-mitchell:
Where were you on this one @coderabbitai?

PR #1007, opened 2025-09-18
FEAT VS-584 (Search) use engine with new categories

jlrhurta:


jcalcines (approved):

jcalcines (changes requested):
@jlrhurta I asume this PR is no longer needed, can it be closed?

PR #1091, opened 2026-02-17
VS-1117-remove-product-search-from-compounds

jcalcines:


jlrhurta (commented):
@jcalcines I guess we are 200% sure this is not in use in any document type (publish or unpublish)

PR #1223, opened 2026-07-09
Vs 1778 se implementarion generate link to google map directions between days

staceywelsh-vs:


site/components/src/main/java/com/visitscotland/brxm/mapper/module/DayMapper.java
```diff
@@ -57,9 +62,32 @@ public ItineraryDayModule map(Day document, PageCompositionHelper compositionHel
...
+        setIntraDayModule(document, compositionHelper);
+
         return day;
     }
 
+    private void setIntraDayModule(Day document, PageCompositionHelper compositionHelper) {
+        try {
+            if (!compositionHelper.getModules().isEmpty()) {
+
+                ItineraryDayModule prevDayModule = (ItineraryDayModule) compositionHelper.getModules().get(compositionHelper.getModules().size() - 1);
```
jcalcines, line 74: suggestion: You should check that the module is instance of ItineraryDatModule.

I know currently only this type of module are allowed but if this changes in the future it will break this item

site/components/src/main/java/com/visitscotland/brxm/model/ItineraryDayModule.java
```diff
@@ -16,6 +16,7 @@ public class ItineraryDayModule extends Module<Day> {
     private FlatLink ctaLink;
     private List<HippoBean> media;
     private MediaSection mediaSection;
+    private IntraDayModule intraDayModule;
```
jcalcines, line 19: nitpick: We don't usually name properties with Module if they are part of another module. 

I think it is ok for the time being but consider removing the word module form the variable name in the future

site/components/src/main/java/com/visitscotland/brxm/services/GoogleMapsService.java
```diff
@@ -107,118 +110,40 @@ public void localizeUrl(FlatLink link, Locale locale) {
...
-
-        TreeMap<Integer, Coordinates> coordinatesMap = new TreeMap<>();
-
-        int dayCount = 0;
-        for (Day dayModule : itineraryPage.getDays()) {
-            final String url = dayModule.getMapLink().getLink();
-            if (url == null) {
-                logger.warn("Null url. Skipping...");
-                return null;
+                logger.info("Could not extract coordinates from previous url");
```
jcalcines, line 127: You should  enhance the logger level to warn and print out the URL so we can troubleshoot the issue if we find it in the logs.

site/components/src/main/java/com/visitscotland/brxm/services/GoogleMapsService.java
```diff
@@ -107,118 +110,40 @@ public void localizeUrl(FlatLink link, Locale locale) {
...
             }
-            Matcher urlMatcher = URL_PATTERN.matcher(url);
+            urlMatcher = URL_PATTERN.matcher(nextUrl);
             if (urlMatcher.matches()) {
-                coordinatesMap.put(dayCount++, new Coordinates(Double.valueOf(urlMatcher.group(1)),Double.valueOf(urlMatcher.group(2))));
+                urlBuilder.append(FWD_SLASH).append(urlMatcher.group(1)).append(",").append(urlMatcher.group(2));
             } else {
-                logger.warn("Url {} did is not a valid google place url.", url);
-                return null;
+                logger.info("Could not extract coordinates from next url");
```
jcalcines, line 135: You should  enhance the logger level to warn and print out the URL so we can troubleshoot the issue if we find it in the logs.

site/components/src/main/java/com/visitscotland/brxm/services/GoogleMapsService.java
```diff
@@ -107,118 +110,40 @@ public void localizeUrl(FlatLink link, Locale locale) {
...
-
-        int dayCount = 0;
-        for (Day dayModule : itineraryPage.getDays()) {
-            final String url = dayModule.getMapLink().getLink();
-            if (url == null) {
-                logger.warn("Null url. Skipping...");
-                return null;
+                logger.info("Could not extract coordinates from previous url");
+                urlBuilder.setLength(0);
+                return urlBuilder.toString();
```
jcalcines, line 129: You should probably return null at this point as an empty URL is not a valid value either

site/components/src/main/java/com/visitscotland/brxm/services/GoogleMapsService.java
```diff
@@ -107,118 +110,40 @@ public void localizeUrl(FlatLink link, Locale locale) {
...
-            Matcher urlMatcher = URL_PATTERN.matcher(url);
+            urlMatcher = URL_PATTERN.matcher(nextUrl);
             if (urlMatcher.matches()) {
-                coordinatesMap.put(dayCount++, new Coordinates(Double.valueOf(urlMatcher.group(1)),Double.valueOf(urlMatcher.group(2))));
+                urlBuilder.append(FWD_SLASH).append(urlMatcher.group(1)).append(",").append(urlMatcher.group(2));
             } else {
-                logger.warn("Url {} did is not a valid google place url.", url);
-                return null;
+                logger.info("Could not extract coordinates from next url");
+                urlBuilder.setLength(0);
```
jcalcines, line 136: You should probably return null at this point

site/components/src/main/java/com/visitscotland/brxm/services/GoogleMapsService.java
```diff
@@ -107,118 +110,40 @@ public void localizeUrl(FlatLink link, Locale locale) {
...
             }
-        }
 
-        for (Coordinates coordinates : coordinatesMap.values()) {
-            if (coordinates == null || coordinates.getLatitude() == null || coordinates.getLongitude() == null) {
-                logger.warn("Null coordinates encountered, skipping...");
-                continue;
+            if (urlBuilder.length() > 0) {
+                urlBuilder.insert(0, DIRECTIONS_URL);
```
jcalcines, line 140: To improve readability you should add this very early when you initialize the urlBuilder.

If you return null after the URLs fail to match you would never get at this point with and empty url

jcalcines (approved):

